PR #3, opened 2018-06-11
[WIP] flesh out basic structure; implement some stuff from CFP C++

msarahan:
Nowhere close to merging.  I'm putting this up primarily for discussion with the VRDI team as they get started tomorrow.

The overall idea is that the program will flow as:

* fileio.py reads in files for initial state.  Returns dataframe.
* models.py will take that dataframe and do some preprocessing.  It will compute the aggregate state of districts and the initial graph.
* metrics.py will define several metrics to be used during the MCMC walk.
* random_walk.py will carry out the iteration.  It will take the aggregate state as input, as well as a list of metrics and optional constraints for those metrics to indicate a bad step.
* The aggregate state will have methods to save its state, either back to files like the ones we have now, or some alternate format.
* The aggregate state will have methods for plotting, or there will be a plots.py file that defines functions that accept the aggregate state and return plots to show.
* api.py will be our only public interface.  We'll refine this as we determine what the really useful, reusable parts of the program are.

### DIFF
--- a/rundmcmc/api.py
+++ b/rundmcmc/api.py
@@ -1,0 +1,6 @@
+"""This file is what external people should interact with.  Be extremely careful in adding
+dependencies here - we want to keep this namespace as simple as possible."""
+
+
+def run_chain(input_file_tuple, steps, *args, **kwargs):
+    pass
--- a/rundmcmc/cli.py
+++ b/rundmcmc/cli.py
@@ -1,3 +1,8 @@
+"""This is the CLI frontend.  It should map to the api as closely as possible.
+
+No functionality should be defined here - just passing through CLI arguments to the API."""
+
+
 def cli():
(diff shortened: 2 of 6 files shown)


msarahan:
It's very impolite to close PRs without any discussion.  I appreciate that ya'll have gone your own direction, but I really don't appreciate tossing my work without any consideration at all.

psward:
That was my mistake on closing.  I saw a request had been waiting multiple days to merge and with how fast we've been moving, I blindly (and dumbly) assumed it was OK to close without checking.  It won't happen again in the future.

msarahan:
No worries, thanks.  If there's useful stuff in here, great.  If not, that's also fine.  It's just important to communicate, and I appreciate your apologies.

PR #6, opened 2018-06-12
Initial commit for underlying graph structure

psward:
Creates the graph object we will be running chain on.  Also includes a way to add data to the nodes of the graph so you can run different objective functions.

### DIFF
--- a/make_graph.py
+++ b/make_graph.py
@@ -1,0 +1,60 @@
+import sys
+import networkx
+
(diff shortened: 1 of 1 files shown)


msarahan:
This was already in https://github.com/gerrymandr/RunDMCMC/pull/3/files#diff-147676499de2f2c59f8f1dc3d3b10c43R4

Is there some way I can coordinate with you all better?

PR #12, opened 2018-06-13
Fixed off by one and districting assignments issues.

psward:


rundmcmc/ingest.py
```diff
@@ -38,15 +35,16 @@ def ingest(filepath, name_of_geoid_col):
...
                 row.append(df["perimeter"][j])
-                adjacency[i][j] = j
+                adjacency[i][j] = int(j)
 
         perims.append(row)
 
     # Strip out zeros from adjacency list.
     for row in adjacency:
-        neighbors.append([i for i in row if i != 0])
+        neighbors.append([i for i in row if type(i) is int])
```
rwbogl: Are we no longer stripping out zeros? If so, then the above comment it incorrect now.

psward: We are still ripping out zeroes.

rundmcmc/make_graph.py
```diff
@@ -14,12 +33,11 @@ def add_data_to_graph(your_data, graph, data_name):
...
     #Adding data to the nodes
     for i,j in enumerate(graph.nodes()):
-        graph.nodes[j][data_name] = your_data[i]
+        graph.nodes[i][data_name] = your_data[i]
 
-def construct_graph(lists_of_neighbors, lists_of_perims, district_list):
+def construct_graph(lists_of_neighbors, lists_of_perims, geoid):
     '''Constructs your starting graph to run chain on
 
     :lists_of_neighbors: A list of lists stating the neighbors of each VTD.
```
rwbogl: Don't forget to update the parameter in the docstring (geoid -> district_list)

PR #13, opened 2018-06-13
Add flake8 linter to circleci build

rwbogl:
This adds a `flake8` linter to the circleci build process. Configuration of the linter can be done in [setup.cfg](/gerrymandr/RunDMCMC/blob/master/setup.cfg) with the already-existing field. Failing code should fail quickly since the linter is run before any of the larger dependencies are installed.


msarahan:
Seems fine to me.  In general, be very careful about mixing conda and pip.  Pip does not respect conda's package constraints (or really even its own), and will break environments.  This is especially true when compiled code is involved.

rwbogl:
@msarahan For future reference, can conda pull Python packages from pypi?

msarahan:
It's a little tedious right now.  You have to create a conda package from it.  There's a tool that's part of the conda-build package that helps: https://conda.io/docs/commands/build/conda-skeleton-pypi.html

We're working on ways to have conda install directly from PyPI, but it's not an easy problem, unfortunately.  Conda requires all dependency metadata up-front, whereas pip finds it iteratively.

PR #17, opened 2018-06-13
Cleanup neighbor selection and documentation in ingest.py

rwbogl:
This is a cleanup of the IO portions of `ingest.py` and `make_graph.py` and an improvement to its documentation. The results of this should be mostly identical to the previous version, including the (probably wrong) perimeter calculations.

I have tested _creating_ the graph from this code, but not using it beyond that.

### DIFF
--- a/rundmcmc/ingest.py
+++ b/rundmcmc/ingest.py
@@ -7,7 +7,10 @@ def ingest(filepath, name_of_geoid_col):
 
     :filepath: Path to input shapefile location.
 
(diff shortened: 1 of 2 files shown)


wtadler:
I was worried that someone would be working on this at the same time as me! See my PR #18.

I'll try to compare the two now.

rwbogl:
To keep the commit history cleaner, I'll just throw the docstring updates in a separate PR.

PR #18, opened 2018-06-13
Compute shared perimeters, streamline ingestion

wtadler:
Previously, ingest was unnecessarily unpacking the pysal weights into an adjacency matrix, but it is conveniently already provided in a dictionary. So I streamlined it.

More importantly, the previous version was not actually computing the shared perimeters between neighbors, but just listing the full perimeter of each shape.

I’ve made it so the output is identical to before (with the corrected shared perimeter values), but I think that the sorting is unnecessary and that the plain neighbors and shared_perims dicts could be handed off to the networkx generation function in their original form, and that integer IDs could be assigned there. Probably doesn’t matter much, but you could potentially have an indexing headache somewhere.

### DIFF
--- a/rundmcmc/ingest.py
+++ b/rundmcmc/ingest.py
@@ -35,4 +35,3 @@ def ingest(filepath, name_of_geoid_col):
 
 if __name__ == "__main__":
     ingest("testData/testData.shp", "CD")
-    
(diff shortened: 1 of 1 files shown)


wtadler:
To avoid the indexing stuff, just take out the `sorted_keys` bit and return the `neighbors` and `shared_perims` dict. There won't be any need for the third output because it's just the keys of the first dictionaries.

Your PR had some docstring work on `ingest` that this could use.

PR #29, opened 2018-06-14
Make ingestion and graph construction super short

wtadler:
This PR shouldn't change any functionality but should improve readability. It also allows the option of keeping nodes labeled by their original GEOIDs for slightly easier matching with other data sources, if you take out the `convert_node_labels_to_integers` line.


wtadler:
I think that the only thing that will change with your output from this PR (if you keep convert_node_labels_to_integers) is that the order of the integer labels assigned by networkx will not necessarily match the order you got before this PR. But the previous ordering was defined by the ordering in whatever shape file you use, and that ordering is not likely to be meaningful.

rwbogl:
@wtadler This PR looks good, but we recently added a unit test for the chain itself. That won't be applied to this PR in a straightforward way. Any objections to changing the base branch to `simplify-ingest` so that we can run that unit test before merging into `master`?

wtadler:
Sure. Do you need me to do anything?

Also, if you all switch away from networkx, then the changes I made in `construct_graph` is probably not useful, but I think the change in `ingest` is still preferable.

wtadler:
Is the fix for the test any more complicated than changing [this line](https://github.com/gerrymandr/RunDMCMC/blob/master/tests/test_updaters.py#L39) in `test_updaters.py` from

```graph = construct_graph(*ingest("rundmcmc/testData/wyoming_test.shp", "GEOID"))```

to

```graph = construct_graph(ingest("./testData/wyoming_test.shp", name_of_geoid_col="GEOID"))```

as I did in this commit in `__main__.py`?

rwbogl:
I merged it into the new branch so that it's easier to run the new tests on it.

> Also, if you all switch away from networkx, then the changes I made in construct_graph is probably not useful, but I think the change in ingest is still good.

The eventual goal is to abstract away the actual graph library implementation, so the `construct_graph` changes might become `construct_networkx_graph` or something like that. Definitely still useful.

> Is the fix for the test any more complicated than changing this line in test_updaters.py from
>
> `graph = construct_graph(*ingest("rundmcmc/testData/wyoming_test.shp", "GEOID"))`
>
>to
>
> `graph = construct_graph(ingest("./testData/wyoming_test.shp", name_of_geoid_col="GEOID"))`
>
> as I did in this commit in `__main__.py`?

That's definitely the first fix. I just tried that and noticed some other problems with the test itself failing. If I had to guess that's probably the test itself and not the changes here.

wtadler:
Okay, cool. Abstracting away the graph library implementation sounds like it's probably essential!

PR #31, opened 2018-06-14
Logging skeleton

apizzimenti:
Provides a functional skeleton for binning statistics as the chain is run. This allows for any stat attached to the partition (e.g. area, population) to be aggregated at a specified interval (which is defaulted to 1% of the total number of iterations). Furthermore, options to save, plot, or output data to console are included. Finally, this doesn't affect the chain's functionality in any way. No functions or data passed to the chain need to be modified to work with the Logger, as they're handled automatically.


apizzimenti:
Will open a new PR – accidentally entered git hell.

PR #32, opened 2018-06-14
ROUGH port to graph-tool

MarshMue:
very first implementation of graph-tool without optimizing to the api

### DIFF
--- a/rundmcmc/make_graph.py
+++ b/rundmcmc/make_graph.py
@@ -1,4 +1,4 @@
-from graph_tool.all import *
+from graph_tool import Graph
 import pandas as pd
 import geopandas as gp
 import numpy as np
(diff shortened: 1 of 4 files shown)


rundmcmc/make_graph.py
```diff
@@ -1,6 +1,7 @@
-import networkx
+from graph_tool.all import *
```
msarahan: Please don't ever do imports this way.  It makes it very hard to understand where functions and variables come from, and it greatly increases the chances of something strange happening if some function or variable clobbers something from another file.

Did this come from a graph_tool example, or what?

MarshMue: The very first code snippet they give 😬
https://graph-tool.skewed.de/static/doc/quickstart.html

msarahan:
much better!  thanks.

MarshMue:
Preliminary profiling has this running slower so far than networkx, which I think is partly due to not utilizing the api as well as it could be. I think we're going to abstract the graph as you suggested before trying to fix those parts, but your input would be appreciated on how to do those speedups! The biggest slowdown came from the merge() method in partition.py, which I'm guessing comes from doing something dumb with the underlying graph structure that is making it quite slow - wasn't much of a bottleneck with networkx.

msarahan:
If your loops are in python, that's why it's still slow.  For example:

https://github.com/gerrymandr/RunDMCMC/pull/32/files#diff-4ad68bcfad290dc257ae02909dbd3c6cR52

You should look for ways to pass larger blocks of data that set all elements at once - that at least provides the opportunity for things to be done in C/C++ in whatever way graph_tool does it.

The major performance killer to watch for is looping in python.  Ask yourself "Do I expect a loop iteration to take more than a few milliseconds?"  If not, try to find a way to do things in bulk.

msarahan:
I ran some profiling on test_updaters.py::test_single_flip_contiguity_equals_contiguity since that is the longest running test right now.  There's good news and bad news.  The good news is that you have done a good job and there's very little time spent in RunDMCMC code - almost all of it is in pysal or networkx.  That's the bad news, too.  There's less low-hanging fruit.  Numba will not be a quick fix.  The speedups will need to be algorithmic, or may need to revert to the more annoying but faster contiguity checks and such from the CFP code.  Those would be better candidates for acceleration with numba or cython.

I think it would be wise to serialize the graph stuff, so that you can load that in without re-running pysal to recompute it all the time.  That will let us focus on the hotspots in the actual MCMC.  Computing it is important - it just belongs in its own test, and it can be serialized and cached to disk for future runs.

PR #36, opened 2018-06-15
try to cache packages for circle ci

msarahan:
per discussion with @rwbogl

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,8 +11,8 @@ jobs:
 
       - restore_cache:
           keys:
-          - v1-dependencies-{{ checksum "setup.py" }}
-          - v1-dependencies-
+          - v2-dependencies-{{ checksum "setup.py" }}
+          - v2-dependencies-
 
       - run:
           name: run linter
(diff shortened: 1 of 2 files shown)


msarahan:
Looks like this cuts time by about a minute and a half.  We might be able to get a little bit faster by switching from mkl to openblas for numpy's blas implementation.

rwbogl:
Wow, that's a big speedup in the dependency stage. (For anyone else reading, cf. this PR's test time with the [previous test](https://circleci.com/gh/gerrymandr/RunDMCMC/106)).

Is there a way to see what's taking up time in the dependency stage post-cache? Reading in the cache itself takes a chunk of time, but then installing dependencies is another minute or so.

msarahan:
It's some combination of downloading the package index plus solving.  You may see a good speedup by pip installing coverage rather than installing that via conda.  There's no particular reason to get it from conda, aside from purity, and having to go to conda-forge to get it makes one more large index to download and do stuff with.

Keep in mind that installing dependencies with conda is just hard-linking stuff that exists in the package cache into the environment.  You still have to pay some price for that, even though the cache saves you the initial download and extract time.

msarahan:
Yep, another 30 seconds.  Beyond this, you start to increase risk with the geospatial libraries.  I don't recommend getting them from pip.  They are very fragile and have intricate binary dependency relationships that conda handles much better than pip.

maxhully:
This is great! Is it time to merge it?

msarahan:
@maxhully now it's ready.

PR #48, opened 2018-06-15
allow csv adjacencymatrix instead of makefile

marybarker:
Still needs some work-outline for the format for adjacencymatrix csv, adding excel option, etc. 
This also includes a change where the added node data can be optionally keyed on GEOID in case there is an ordering issue that needs to be resolved.


msarahan:
It looks like you might benefit from using an editor that is more python-aware.  For example, mine keeps translates between tabs and spaces (keeping only spaces, but I mostly only ever hit tab).  I use spacemacs, and that's pretty esoteric.  The free community edition of pycharm or visual studio code might be worth your time to install and learn.  These will also have live style-checking so you don't get hung up on it at test time on the CI's.

maxhully:
I have no expertise in vim, but based on my googling it looks like there are plugins for autopep8 and flake8. Might be worth a look!

PR #52, opened 2018-06-16
Hopefully more efficient cut_edges

maxhully:
I implemented a different version of cut_edges that should hopefully be more efficient. Before this gets merged, we should talk about:

1. I implemented cut_edges as an updater. Is this a good idea? The proposal function depends on it being there, and so far nothing else does. Does this mean it should be a method of Partition? Should the proposal be a method of Partition, and then we'd create new proposal functions by providing alternate Partition implementations/subclasses?

2. Now cut_edges returns a set, because it was easier to write that way. Are sets bad?


msarahan:
No idea on 1 - that's more than I know about for the design now.

For 2, sets are not intrinsically bad, but they do have characteristics to keep in mind:

- they are not ordered.  If you turn a list into a set, its order is destroyed
- they have very fast lookup (hash map), so things like checking if an element is in the collection will be much faster with a set than a list
- Sets are slower than lists for iterating over members
- You cannot access a set element by numeric index

https://stackoverflow.com/questions/2831212/python-sets-vs-lists

maxhully:
Hmm... For computing which edges cross parts I don't have to iterate over cut_edges, but the proposal function does a random.choice of the cut_edges. That's okay, because we will eventually want to scrap that and use the pre-generated random numbers like you (Mike) wrote in your pull request.

We could use a randomly generated number between 0 and the total number of edges, and then return the corresponding edge if it is in the set of cut_edges, otherwise grab the next random number off the stack and repeat.

msarahan:
That's kind of what I was thinking with the big pool of random numbers, and I think it's worth a try.  I have no idea whether that would be any faster, though.  Sure, generating the random numbers will be faster, but if it is only using a small fraction of the edges, it seems like it might be slower.  If you could have the pool pick from only the cut_edges set, that would be faster (but then you're indexing a set, which means you'd have to convert it to a tuple or list).

maxhully:
Yeah, picking only from cut_edges would be ideal. But I'm not sure how we would generate the random numbers ahead of time then, since the size of cut_edges will vary over the life of the chain.

PR #57, opened 2018-06-18
Latest version of the loggers

maxhully:
I'm starting to wonder if I pushed us into some premature abstraction with these loggers...


maxhully:
#68 is the more up-to-date version.

PR #63, opened 2018-06-20
added MO data

marybarker:
just some shapefiles so we can play with data.


rwbogl:
At some point, maybe our test data should be in the tests folder, rather than with the source code. (No need to quibble about it right now though.)

PR #65, opened 2018-06-20
Score functions and pickled the graph!

maxhully:


msarahan:
I don't know about networkx's pickle implementation specifically, but in general, pickles are poor for doing much of anything permanent.  They are not guaranteed to be compatible across python versions.  I highly recommend considering one of the other options.

maxhully:
Good to know! The pickle is temporary so that people can play around with the chain this afternoon without needing to load the data by hand. I'll make sure not to use it for anything long-term.

msarahan:
Cool.  As long as everyone is using the same python version, you should be fine.

rwbogl:
Changing the base branch to `sandbox` so we can sort through this PR after playing around this afternoon.

PR #91, opened 2018-06-26
removed uneccesary cut_edges call

Untellable:
This speeds things up bc it removes an unecessary call to cut_edges.


psward:
Great find!

PR #93, opened 2018-06-27
PA data and basic partition set up

marybarker:
- PA shapefiles and adjacency graph with data 
- cleaned up graph creation process so we only need one function call to create a networkx graph and add data to it (in make_graph all of the construct_graph_* functions)
- modified Wes adjacencies to finally (hopefully) fix the off-by-one problem


maxhully:
Woo!

PR #97, opened 2018-06-27
Main script for PA runs

maxhully:
The __main__ script now outputs some nice histograms. Woo!


maxhully:
Also contains some of the new proposal functions from the proposals group (woo!).

PR #107, opened 2018-06-29
Config file format added

marybarker:
CHANGES:
- Added config file parser so that main does not have to change for every run
- Changed get_list_of_data in make_graph to return a dataframe rather than a list of tuples

NOTES:
- basic config file sets up and runs PA for 1000 steps with histogram output per previous commit
- config parser file is MESSY. TODO: clean, standardize
- GUI for config file generating will be added soon to simplify runs

TODO: 
-  add the functions that visualize output to run.py (and make more. e.g. plots as well as hists)
-  add dependency list for updaters based on the validity/score functions that are selected

### DIFF
--- a/rundmcmc/__main__.py
+++ b/rundmcmc/__main__.py
@@ -10,5 +10,6 @@ def main():
 
     output_func(output, scores)
 
+
 if __name__ == "__main__":
     main()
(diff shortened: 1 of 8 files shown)


marybarker:
from last TODO: 

- a manual list of updaters dependencies is now in place, with code that checks validators and scores to automatically populate updaters. Probably need to have a conversation about how to do this in a prettier way in future. 

- added viz functions to output.py (Still need to standardize this form though)

PR #112, opened 2018-07-02
High-performance data storage

apizzimenti:
The `DataStore` class allows users to store data efficiently. This object can be iterated over, indexed, stringified, and `.append()`ed to. All operations are subject to change, but as it stands, this is a time- and space-efficient way to store massive amounts of data.


rwbogl:
Only fails tests involving matplotlib. That seems to be a systematic failure for now. I'll look into fixing up our circleci config to account for the new import.

PR #113, opened 2018-07-02
Memory and performance improvements

rwbogl:
This PR makes two minor changes that give major improvements.

First, Partition objects only track one level of parents. The chain manually deletes anything further back. This lets Python's garbage collection get at the old Partition instances, freeing up HUGE amounts of memory. (On 50,000 steps, I went from 2 GB to 10 MB.)

Second, `propose_random_flip` has been reorganized so that it exits early when possible. This gives a minor performance boost depending on the underlying graph.

__EDIT:__ It also fixes the failing test issue by changing the matplotlib import. That might break something else unintentionally.


maxhully:
Woo! Way to pass the tests!

PR #125, opened 2018-07-09
Optimized (hopefully) Polsby-Popper.

maxhully:
- Changed Polsby-Popper to use the add/subtract trick instead of just summing everything up.
- Renamed `polsby_popper_updater` to `polsby_popper` after all.
- Split updaters up into multiple files within its own package.
- Added a test to check that the Basic Chain with `PA_partition` runs.


mahrud:
This PR broke `setup.py`, but #158 fixes it.

PR #133, opened 2018-07-10
run chain with rundmcmc config.ini

hwheelen:
* Can run chain outside of rundmcmc folder
* Can output directly into the folder you want
* Don't need to call `__main__` directly. Instead `rundmcmc config.ini`

### DIFF
--- a/rundmcmc/__init__.py
+++ b/rundmcmc/__init__.py
@@ -1,3 +1,4 @@
+from .__main__ import main
 from ._version import get_versions
 
 __version__ = get_versions()['version']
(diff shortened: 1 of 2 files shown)


mahrud:
Adding ```assert main``` at the end of `__init__.py` will silence the CI test.

mahrud:
Alternatively, if you change `__main__.py` like this: (notice changes in lines 1, 2, and the last line)
```py
def main(args = sys.argv):
    chain, chain_func, scores, output_func, output_type = read_basic_config(args[1])
    print("setup the chain")

    output = chain_func(chain)
    print("ran the chain")

    output_func(output, scores, output_type)


if __name__ == "__main__":
    main()
```

Then after installing the module, from any directory you can run `rundmcmc config.ini` and :boom:.

Suggestion: make a way to overwrite `num_steps` from command line!

maxhully:
Woo! First pull request!

PR #139, opened 2018-07-10
Add a useful instruction.

apizzimenti:
If people want to make edits to the repo (which they shouldn't have to but inevitably will), they have to run `python setup.py develop`; otherwise, their changes aren't reflected in the code they run. This is a basic first step, but I also propose that we _only_ include the `develop` step. It does exactly what `install` does, but allows for changes to be made.


rwbogl:
> I also propose that we only include the develop step. It does exactly what install does, but allows for changes to be made.

I agree; if we're advocating for `develop`, then it's silly to run `install` right before it. (At least, I don't _think_ there's any benefit to it.)

rwbogl:
Like you mentioned, in the "future," this shouldn't be necessary. For the time being, it's the easiest fix for people that are hacking around inside the source directories.

PR #151, opened 2018-07-11
Extend IO to support all fiona drivers

rwbogl:
This touches up the graph construction functions to support everything that `fiona` can handle. Before, we were artificially limiting ourselves to JSON and shapefiles. Now, we have shapefiles, GeoJSON, and a whole slew of other GIS stuff I don't understand.

Here's a functional change: There is no longer a method to load a graph from a JSON string. All JSON graphs are loaded from files. I don't think anyone actually used this functionality yet, and it's easy to add back in if we want it.


rwbogl:
Notes for later: This PR won't play nice with the GUI if people drop in NetworkX JSON graphs. The file will get passed to `construct_graph()`, which now by default assumes that the object is a file to pass off to geopandas. This means that the GUI needs to know if it's reading in something "like" a shapefile or a NetworkX graph. I don't think that anyone is using the NetworkX JSON feature right now, but we should figure this out before merging.

rwbogl:
Here's my proposed fix: #169 added a `save_json` option to the config file generated by the GUI. If a user wants to load in a JSON graph, then they have to use that option. The `gSource` option must point to some fiona-supported geography file (shape, GeoJSON, etc.). This will require manually editing the config file if you want to load a JSON graph in, but if you were using that feature then you probably already know how to do that.

@drdeford I fixed your template files here. Thoughts on the changes/this fix?

PR #156, opened 2018-07-11
Added simple accepts and un-hardcoded plots

drdeford:
Added 2 simple accept functions and 1 metagraph proposal. 

Changed the histogram plot to take number of bins as input and changed pipe_to_table to have separate bin vs. plot intervals. 

Failed test is due to the random import in proposals.py.

### DIFF
--- a/rundmcmc/template_main.py
+++ b/rundmcmc/template_main.py
@@ -91,7 +91,7 @@ acceptance_method = always_accept
 
 
 # Number of steps to run
-steps = 100000
+steps = 1000
 
 print("loaded data")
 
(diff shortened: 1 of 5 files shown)


rundmcmc/template_main.py
```diff
@@ -0,0 +1,189 @@
...
+area_col = "ALAND10"
+district_col = "CD"
+
+
+# This builds a graph
+graph = construct_graph(graph_path)
+
+# Write graph to file
+with open('graph_with_data.json', 'w') as outfile1:
+    outfile1.write(json.dumps(json_graph.node_link_data(graph)))
```
rwbogl: This should be `json_graph.adjacency_data(graph)`, I think. I can't remember why, but I had issues when using `node_link_data()`. Maybe someone can chime in and explain.

drdeford: Changed!

PR #158, opened 2018-07-12
makes 'python setup.py install' install the rundmcmc.updaters package as well

mahrud:
Also fixes the `rundmcmc PA.ini` routine.


rwbogl:
Thanks for the fix!

PR #171, opened 2018-07-12
fixed output for json

marybarker:


mjharris95:
still doesn't seem to be working correctly.

here's the beginning of the json now: {0: null2: null3: null4: null5: null6: null7: null8: null9: null10: null11: null12: {"0112514": "06"}13: {"0112514": "06"}

PR #184, opened 2018-07-17
Overhaul validator style and docs

rwbogl:
This does two things: Add docs to every possible function in `validity.py`, and order functions in a more cohesive way. (By "cohesive," I really just mean that classes come first and a few helper functions are now inline/hidden by underscores.)


drdeford:
This is great!! Exactly the kind of clean, informative setup that will be helpful for contributors.

PR #187, opened 2018-07-18
[WIP/RFC] Project restructuring

rwbogl:
This is a restructuring of the project into a collection of subpackages rather than just flat modules. I don't _think_ that this breaks anything, but just to be safe I suggest merging this into a separate branch when we get to it.

This PR is so people can easily view and comment on the changes (what should go where, `__init__.py` practices, etc.) before things get too far along. Suggest changes!

For quick reference, here's what the new structure looks like:

tree


rwbogl:
So that this PR doesn't lag too far behind current changes, if there are no objections then I'll merge this tonight or tomorrow morning after fixing the circleci issues.

maxhully:
Sounds good!

PR #190, opened 2018-07-20
Fix add_boundary_perimeters to touch all nodes

rwbogl:
Now all nodes should have the `boundary_node` boolean attribute. Sorry for the confusion, @drdeford!


drdeford:
All good! I like the new consistent df setup.

PR #195, opened 2018-07-23
Fix cut_edges_by_part

maxhully:
- Fix `cut_edges_by_part`, which fixes perimeters/Polsby-Popper.
- Add more tests for `cut_edges_by_part` to show that it responds properly to flips. Even in the areas where we have test coverage, we could use more testing of how things respond to flips.
- Fix Mean-Thirdian to be (the third lowest R%) - (the mean R%). It's now giving more sensible results, in that it no longer sees absolutely every plan as an extreme D gerrymander.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,5 @@ junit.xml
 
 # Pytest cache
 .pytest_cache/
+
+Outputs/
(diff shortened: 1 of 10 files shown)


maxhully:
About the terminology: agreed. I've come to regret the `cut_edges` name. I think Robert has some thoughts about ways to make the API more intuitive too.

maxhully:
I'll merge this in so that people can pull it down at 10am.

PR #199, opened 2018-07-24
Add warning messages to make_graph.py

rwbogl:
This also changes the `geo_data_frame` data source type to `geodataframe`, which is more consistent with the actual name. (Though maybe it should be `GeoDataFrame`.)

### DIFF
--- a/rundmcmc/make_graph.py
+++ b/rundmcmc/make_graph.py
@@ -1,6 +1,7 @@
 import networkx
 import pandas as pd
 import geopandas as gp
+import warnings
 import pysal
 import json
 from networkx.readwrite import json_graph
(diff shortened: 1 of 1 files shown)


rundmcmc/make_graph.py
```diff
@@ -136,14 +137,21 @@ def construct_graph_from_df(df,
...
         pops = df[pop_col].to_dict()
+    else:
+        warnings.warn("No population column was given, assuming all 0")
 
-    areas = df['geometry'].area.to_dict()
     if area_col:
         areas = df[area_col].to_dict()
+    else:
+        # This may be slightly expensive, so only do it if we know we have to.
+        areas = df['geometry'].area.to_dict()
```
ljwolf: fyi, geopandas generally recommends using `geodataframe.geometry` to refer to the active geometry column, since that is [always the active geometry column](https://github.com/geopandas/geopandas/issues/755).

PR #200, opened 2018-07-25
Correct import

### LINKED ISSUE #130: Misleading import
In defaults.py, propose_random_flip_no_loops is imported as propose_random_flip, which is a different function which does have loops. Is there a particular reason it wasn't just imported under it's full name?

maxhully:
We recently set `propose_random_flip = propose_random_flip_no_loops`, so now this won't change anything.

Closes #130 .

### DIFF
--- a/rundmcmc/defaults/defaults.py
+++ b/rundmcmc/defaults/defaults.py
@@ -8,8 +8,7 @@ from rundmcmc.chain import MarkovChain
 from rundmcmc.make_graph import (add_data_to_graph, construct_graph,
                                  get_assignment_dict_from_df)
 from rundmcmc.partition import Partition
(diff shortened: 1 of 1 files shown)


drdeford:
In proposals.py I think it is still propose_random_flip_no_loops? (although I would be up for renaming and then calling the other one propose_random_flip_with_loops)

maxhully:
Let's do that. I'll make the change.

maxhully:
Actually I can't because I'm doing this through the GitHub UI on my phone...

drdeford:
Looks good, we'll just have to warn people about changing their proposal import.

maxhully:
Thanks @rwbogl!

PR #203, opened 2018-07-25
WIP: Basics for the initial report construction

drdeford:
Added an initial report function and a template that processes the data and then calls initial report to build the html output. Currently set up to run against as many plans and election sources as you have on your shapefile and needs an associated geojson for plots. Added a couple of other simple scores for simplicity. 

Very much a work in progress but provides a starting point for beautifying. 

Will undoubtedly fail lots of tests.


rwbogl:
> Will undoubtedly fail lots of tests.

:tornado:

(I'm disappointed that there isn't a "torch and pitchfork" emoji.)

The outputs here look great though!

drdeford:
Yeah, 🔥 . The problem is that writing html with python this way is a little clunky and splitting up the lines nicely is a little unintuitive.

maxhully:
One strategy I'm aware of is templating packages like http://jinja.pocoo.org , although that would require learning how to use that...

drdeford:
Newest version includes a county level entropy report as optional. 
It fails even more tests.

msarahan:
I can help with Jinja2.  It's very useful and pretty easy to learn.

Just write your html as normal, but you then use extra syntax to leave placeholders for stuff from python.  Jinja2 fills those in.

Jinja2 won't generate latex for you, but it works on any kind of template file.  Think of it as a text preprocessor.

drdeford:
That would definitely be helpful! This approach works but it feels pretty kludgy.

msarahan:
Go ahead with this, and I can help you change it over to jinja2 stuff later.  There are other tools, such as sphinx, that let you generate "docs" from python code plus some markdown (traditionally RST, but I think markdown works these days).  That then can generate html, latex, PDF, or many other outputs.  I recommend an issue to talk through exactly what the goals are, and then I can help you achieve them.

PR #207, opened 2018-07-27
Forces columns on dataframes to be numbers

drdeford:
Some shapefiles were giving populations, votes, etc. as strings. Now make_graph coerces those to be numeric on the dataframe.


drdeford:
Now we are not renaming columns if we were given a name to begin with for area, population, and district. Uses the defaults if they don't.

drdeford:
Now we are no longer overwriting column names for pop_col, area_col, and most importantly district_col if they are provided to construct_graph.

PR #208, opened 2018-07-30
fixes TypeError when reading field types in tally.py

hwheelen:
This PR fixes the same problem as #207, but in a different way.


rwbogl:
Do we need both this and #207? If I'm not mistaken, #207 should convert everything to a numeric type (probably float) that it can, meaning that the fix here isn't always needed.

PR #209, opened 2018-08-03
L1 and L2 Scores

drdeford:
Short addition to validity to allow L1 and L2 scores for MA. Also updated init to expose more functions.


drdeford:
Oops, branched from the wrong spot. Closing this one and reopening.

PR #211, opened 2018-08-06
Remove old testData

drdeford:
Remove the old data to simplify folder and store fewer shapefiles. Will replace with 1 PA file for testing. The old data can be downloaded from people.csail.mit.edu/ddeford/data_code.php


drdeford:
Removind testData makes circleci sad.

Issue #103, opened 2018-06-28
Allow checking if yielded state was accepted


rwbogl:
Currently, the chain yields all valid proposed states. There is no way for a user to check if a state yielded by the chain iterator was accepted. (We know if it's _valid_, but not accepted.)

maxhully:
(We resolved this on slack)

rwbogl:
I believe that the conclusion on Slack was that the chain doesn't do this. I think that it would be worthwhile to add this, as otherwise users must duplicate the work of the acceptance function in their chain iterations.

maxhully:
Added this in #214 .

PR #214, opened 2018-08-08
Raise a TypeError if a constraint returns non-boolean

maxhully:
This will catch the bug that we noticed a couple weeks ago where a constraint (in particular `refuse_new_splits`) is configured improperly and ends up returning a non-boolean value that resolves as truthy, making it seem like the partition passes the constraint when really it isn't.

Adding this check showed that BasicChain has this problem (wrongly configured `refuse_new_splits`), so I removed that constraint from BasicChain.

This new check will cause errors anywhere that we still have the wrong usage of `refuse_new_splits` hanging around. I think this still includes all of the config stuff. **We should talk about if we should wait to merge this until after Your State 2.0.**

I also added some tests assuring that `MarkovChain` returns the initial state in the first step and that `pipe_to_table` records the initial values in the first row.


maxhully:
(I have some old commits from that ChainMap concept in here, but it didn't work so I took them back out.)

drdeford:
Nice! state_report and generate_ensemble are the your state 2.0 files and neither of them is currently using refuse_new_splits although we are calculating county split entropies so most (much of, some, a few) of the runs will probably have well-behaved county data that we could use.

rundmcmc/chain.py
```diff
@@ -54,15 +54,20 @@ def __next__(self):
                 else:
                     continue
 
-            self.state.parent = None
+            # Please delete this comment
```
rwbogl: :2nd_place_medal:

maxhully: Lol! That was from my "editing on my phone in the GitHub editor" phase, when I could create new lines but not delete any

maxhully:
Also adding an `accepted` boolean field to all states that the chain yields. This closes #103 .

PR #231, opened 2018-10-16
Change link to getting started guide

katiejolly:
The link that was there was broken, so I found the link I think it's supposed to point to.


rwbogl:
I agree, this looks like the right link. Thanks for fixing it!

maxhully:
Congrats on your first pull request! :tada:

PR #248, opened 2018-12-06
Some simple methods for recursively building new seeds

drdeford:
This contains a wrapper for turning any bipartitioning method into a new seed generator for any number of districts and two new bipartitioning methods. The first uses a minflow algorithm to cut the graph and the second randomly deletes edges until a balanced partition is found.


drdeford:
This will be replaced by week 6 vrdi 2019.

drdeford:
Superseded by: https://github.com/vrdi/seeds

PR #250, opened 2018-12-07
ReCom proposal

maxhully:
This PR merges in a small part of the material from #242 , covering only what is needed for the ReCom proposal from the 2018 Virginia House of Delegates report.

Also includes a fix for readthedocs, and makes `Assignment` callable so that we can use it for making plots with geopandas.


maxhully:
Another note: `MarkovChain` now takes a `constraints` argument, which can either be a callable or a list of constraints (callables). If it's a callable (like `Validator`), that's used as `is_valid` like before. Otherwise we instantiate the validator from the list of constraints.

This will break code that uses `is_valid` as a keyword argument (but not as a regular argument).

maxhully:
A possible simplification for `recom`: if we make a new population constraint class that contains the ideal population (`pop_target`) and the `epsilon` value (and maybe even `pop_col`?), then we could use that object instead of those 2 or 3 arguments when calling `functools.partial`.

Issue #241, opened 2018-11-29
Partisan symmetry scores


mduchin:
It would be great to add 2 other symmetry scores to supplement the mean-median:  we can do partisan bias (the number of districts with above-mean Republican share divided by the total number of districts, minus 1/2) and partisan Gini (the area between the seats-votes curve and its reflection over (.5,.5))

maxhully:
Thanks to @pjrule , we now have this!

PR #261, opened 2019-01-06
Partisan bias and partisan Gini scores

pjrule:
This PR implements issue #241. It adds two new scores to `gerrychain.scores`:
- **Partisan bias.** The scoring function implements @mduchin's definition ("the number of districts with above-mean Republican share divided by the total number of districts, minus 1/2") but uses the first party of the given `ElectionResults` object rather than hardcoding the GOP. This follows the convention used by the other scoring functions.

- **Partisan Gini.** This score is rather esoteric—there seems to be little definitional information available other than #241. I referred to [these Tufts Math 19 lecture notes (pdf)](https://sites.tufts.edu/socialchoice/files/2018/01/partisan.pdf), which explain how to generate a seats-votes curve from a single election under the assumption of uniform partisan swing. The area between this curve and its reflection about (.5, .5) is calculated in accordance with @mduchin's definition of the metric. In this plot, the area of the yellow region gives the partisan Gini score:

pa_seats_vs_votes

Unit tests are included; some previously existing unit tests are updated to include the partisan bias metric. Unlike all other currently defined metrics, partisan Gini is unsigned, so it is not included in any of the existing tests related to directional checks.

### DIFF
--- a/gerrychain/scores.py
+++ b/gerrychain/scores.py
@@ -70,9 +70,9 @@ def partisan_bias(election_results):
     :election_results: the ElectionResults to compute partisan bias against.
     :party: the party to compute partisan bias against.
     """
-    first_party  = election_results.election.parties[0]
+    first_party = election_results.election.parties[0]
     party_shares = numpy.array(election_results.percents(first_party))
-    mean_share   = numpy.mean(party_shares)
+    mean_share = numpy.mean(party_shares)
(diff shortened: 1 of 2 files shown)


maxhully:
Looks awesome! Thanks for implementing this, and congrats on your first PR! 🎉

PR #262, opened 2019-01-10
[Bug?] Invalid geometries after reprojection

pjrule:
As a new GerryChain user, I wanted to run GerryChain on [MGGG's official Wisconsin shapefiles](https://github.com/mggg-states/WI-shapefiles). However, I was initially unable to load these files with `Graph.from_file()`—I was met with a fatal `TopologyException` related to self-intersection. I discovered that turning reprojection off using `Graph.from_file(reproject=False)` alleviated this error; further manual inspection using GeoPandas indicated that reprojection can introduce self-intersections into previously valid shapefiles. One [common fix](https://stackoverflow.com/a/20873812) for these issues, which I've used in previous GIS experiments, is to call `.buffer(0)` on the offending geometries. From the [Shapely user manual](https://shapely.readthedocs.io/en/stable/manual.html#constructive-methods):
> Passed a distance of 0, `buffer()` can sometimes be used to “clean” self-touching or self-crossing polygons such as the classic “bowtie”. Users have reported that very small distance values sometimes produce cleaner results than 0. Your mileage may vary when cleaning surfaces.

I've implemented a proof-of-concept validation loop in `gerrychain.graph.geo` to partially address this bug using the `.buffer(0)` method. There are still some subtleties to be addressed—modifying the edges of a precinct, even minutely, tends to cause overlap with other precincts. **Update: we've decided this isn't the best idea and that we'd rather raise a helpful error when invalid geometries occur. See thread below.**

It's quite possible this is merely an issue with the particular shapefile—I was unable to reproduce the bug for other MGGG shapefiles—but I suspect that this issue extends beyond the shapefile in question and that it may be appropriate to address this issue directly in GerryChain. Thoughts? 😀

#### Reproducing the bug
```python
>>> from gerrychain import Graph
>>> g = Graph.from_file('WI_ltsb_corrected_final.shp')
```
Output:
```
TopologyException: Input geom 0 is invalid: Self-intersection at or near point 237345.77388761958 4847210.1287339805 at 237345.77388761958 4847210.1287339805
/Users/pjrule/GerryChain/gerrychain/graph/geo.py:37: UserWarning: Repaired invalid geometry at index 2266 after reprojection.
  "reprojection.".format(idx))
TopologyException: Input geom 1 is invalid: Self-intersection at or near point 237345.77388761958 4847210.1287339805 at 237345.77388761958 4847210.1287339805
---------------------------------------------------------------------------
TopologicalError                          Traceback (most recent call last)
<ipython-input-2-91a4f7860563> in <module>()
      1 from gerrychain import Graph
----> 2 g = Graph.from_file('WI_ltsb_corrected_final.shp')

~/GerryChain/gerrychain/graph/graph.py in from_file(cls, filename, adjacency, cols_to_add, reproject)
     61         """
     62         df = gp.read_file(filename)
---> 63         graph = cls.from_geodataframe(df, adjacency, reproject)
     64         graph.add_data(df, columns=cols_to_add)
     65         return graph

<deeper stack trace omitted>

TopologicalError: The operation 'GEOSIntersection_r' could not be performed. Likely cause is invalidity of the geometry <shapely.geometry.multipolygon.MultiPolygon object at 0xa253a9358>
```

#### Versions
- GerryChain: `master` (`5e50016`)
- GeoPandas: 0.3.0
- Shapely: 1.6.4.post1


zschutzman:
For the future, maybe it would be good if GerryChain did something more useful than printing a geopandas stack trace and quitting. A `try/except` block could be used to try to do something else, like attempt to build the graph with `reproject=False` or call `.buffer(0)` and try again.

maxhully:
I think my preference would be to recommend `.buffer(0)` as a possible fix in the error message, but still raise the error. I don't think we should automatically try to fix the issue unless asked to do so. (Although, we are already reprojecting by default, so maybe I'll have to think harder about this philosophy... Maybe the behavior should be to reproject if the geometries are in long-lat coordinates and issue a warning, and otherwise leave it be?)

I also want to know more about how big of an effect `.buffer(0)` might have on the geometries, and whether it is likely to work as a fix for all the possible `TopologicalError`s that might show up.

For the Wisconsin-specific CRS issue, it sounds like maybe the method we're using to pick the UTM zone could be improved.

zschutzman:
Somebody with more geography knowledge than me should tell me that I'm very wrong about this, but since the US is in the northern hemisphere, it's never "wrong" to use the 'N' version of the WGS84 UTM zone over the regular no-letter one.  The function which finds the most frequent UTM number also finds the right letter, but that information isn't being retained as of right now.

Also agreed that the default behavior should be an informative error/warning rather than trying to fix a user's geometry.  What `.buffer(0)` does is return the 16-gon envelope approximation of your shape. This seems like a bad choice when working with geometries that are as complex as these geographies can be. You can pass a second argument `.buffer(dist, verts)` where `dist` tells you how far out you want your envelope and `verts` is the number of vertices to use. If we want to use it or recommend the user to use it, we should think about the best way to make sure we use enough points to keep the important geometric information. [The source for `.buffer()` is here.](https://github.com/Toblerity/Shapely/blob/0d72c3d63802c9d16cf235bb5166b11384730226/shapely/geometry/base.py)

zschutzman:
Okay, I did some more digging to see if I could figure this thing out. In the WI file, there is a ward (GEOID10 = 55057885250002) which is a multipolygon, composed of three pieces. There is a "main" piece, a small triangle off to the southwest, and a sliver that's part of the main piece
reproj_bad_whole

When we call `reprojected()`, it does math on each piece separately, and the sliver crosses over into the main piece probably just as a result of the computer doing floating point arithmetic.
badzoom

Separately, `.buffer(0)` is somewhat of a folk remedy in the geospatial analysis world.  It's not designed to correct these geometric errors and there's sort of no way of knowing when it will or will not resolve the problem. I don't think we should include something with poorly defined behavior in GerryChain.

pjrule:
> Separately, .buffer(0) is somewhat of a folk remedy in the geospatial analysis world. It's not designed to correct these geometric errors and there's sort of no way of knowing when it will or will not resolve the problem. I don't think we should include something with poorly defined behavior in GerryChain.

Indeed—`.buffer(0)` definitely has some undesirable side effects, some of which are observable in this specific case. I put this code up because I'm using it as a temporary hack and I figured it might be a useful jumping-off point. 😉

From a usability perspective, I like @maxhully's idea of raising a specific error that possibly mentions the `.buffer(0)` trick—it addresses topological errors more gracefully than the current implementation, but doesn't automatically sacrifice correctness.

pjrule:
@maxhully Sounds great! I'll do the implementation and try to get it up within a day or so. :)

pjrule:
Shapely [includes a function to validate geometries](https://github.com/Toblerity/Shapely/blob/master/shapely/validation.py). I'm thinking of using it for the validation and doing something like this:

```python
def reprojected(df):
    utm = identify_utm_zone(df)
    reproj_df = df.to_crs(
        f"+proj=utm +zone={utm} +ellps=WGS84 +datum=WGS84 +units=m +no_defs"
    )
    for idx, row in reproj_df.iterrows():
        validity = explain_validity(row.geometry)
        # This specific string is used by GEOS/Shapely to denote valid geometries.
        if validity != "Valid Geometry":
            raise ProjectionError(idx, validity)

class ProjectionError(Exception):
    def __init__(self, index, error):
        msg = "Invalid geometry at index {} (error: {}). Consider reloading the "
              "shapefile with `reproject=False` or attempting to repair the affected "
              "geometry with .buffer(0)`.".format(index, error))
        self.message = msg
```

With regards to the `.buffer(0)` trick, I've noticed an issue with this approach. In most cases, it's necessary to buffer _after_  reprojection. Throwing an error during reprojection makes this impossible using GerryChain's API alone—to apply `.buffer(0)`, a user would need to load the map using GeoPandas in its original projection, reproject it with GeoPandas, fix the affected geometry, and _then_ pass the map to GerryChain. For this reason, I'm thinking of dropping the buffering message altogether and retaining only the `reproject=False` suggestion—integrity issues aside, it may not even be a particularly usable fix.

zschutzman:
The `explain_validity()` loop is fairly low overhead -- it may be worth checking the geometries even if `reprojected()' isn't called.

It's a pretty safe assumption that a shapefile downloaded from some random state's data repo will be full of errors, and the more info we can give the user, the better, since the default error messages are pretty useless.

maxhully:
This looks great! I agree with Zach---we should do these checks for every geodataframe. We could extract the `explain_validity` loop as its own function, and then call it before and after reprojecting in `Graph.from_geodataframe`?

pjrule:
@maxhully Yeah, that's what I was thinking based on @zschutzman's comment. I'll prototype something. :)

maxhully:
Let's merge this in! Sorry for leaving you hanging @pjrule .

pjrule:
No problem. Thank you for the guidance! 😁

Issue #280, opened 2019-03-12
Require assignments to match the nodes of the graph


maxhully:
If the user tries to instantiate a Partition with an assignment whose keys don't match the nodes of the graph, we should raise an exception.

zschutzman:
I think adding the following method to `Partition` 

```
    def validate_assignment(self):
	node_names = set(list(self.graph.nodes()))
	assgn_names = set([name for dist in self.assignment.parts.values() for name in dist])
	return node_names == assgn_names
```

and calling it in `Partition._first_time()`

```
    def _first_time(self, graph, assignment, updaters):
        self.graph = graph

        self.assignment = get_assignment(assignment, graph)

        if not self.validate_assignment():
             raise NameError("Graph's nodes' names do not match the Assignment's geo units' names.")

        if updaters is None:
            updaters = dict()
      ...
```

does the trick. I'm happy to do the pull request if you think this is a good way to solve it and this is the correct exception to raise.

maxhully:
This is making me think that `Assignment` should implement the `Mapping` interface in [collections.abs](https://docs.python.org/3/library/collections.abc.html) (or at least have a `.keys()` method that iterates through all the names in the parts). If we do that, then I think we can inline `validate_assignment` to `set(assignment) == set(graph.nodes)`. For the error type, maybe `KeyError` or just plain old `ValueError` would be more appropriate?

How about you open a PR, and we continue this discussion there?

maxhully:
Closed by #281 . Thanks Zach!

PR #281, opened 2019-03-13
first implementation of validating node names

zschutzman:
Migrating discussion of Issue #280 (making sure assignments have nodes named in the same way as the graph).

Points to discuss: 

- Is the initialization of the `Partition` object the right place to do this check?
- Should the `Assignment` class implement the `collections.Mapping` stuff?
- Is `NameError` the right exception? `KeyError`? `ValueError`?

### DIFF
--- a/gerrychain/partition/partition.py
+++ b/gerrychain/partition/partition.py
@@ -56,7 +56,7 @@ class Partition:
         self.edge_flows = None
 
     def validate_assignment(self):
-        node_names = set(list(self.graph.nodes()))
+        node_names = set(self.graph.nodes)
         assgn_names = set([name for dist in self.assignment.parts.values() for name in dist])
         return node_names == assgn_names
 
--- a/gerrychain/partition/partition.py
+++ b/gerrychain/partition/partition.py
@@ -57,7 +57,7 @@ class Partition:
 
     def validate_assignment(self):
         node_names = set(self.graph.nodes)
-        assgn_names = set([name for dist in self.assignment.parts.values() for name in dist])
+        assgn_names = set(name for dist in self.assignment.parts.values() for name in dist)
         return node_names == assgn_names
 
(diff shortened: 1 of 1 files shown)


gerrychain/partition/partition.py
```diff
@@ -51,6 +55,11 @@ def _first_time(self, graph, assignment, updaters):
         self.flows = None
         self.edge_flows = None
 
+    def validate_assignment(self):
+        node_names = set(list(self.graph.nodes()))
```
maxhully: I think we can shorten this to:
```suggestion
        node_names = set(self.graph.nodes)
```

gerrychain/partition/partition.py
```diff
@@ -51,6 +55,11 @@ def _first_time(self, graph, assignment, updaters):
...
+        assgn_names = set([name for dist in self.assignment.parts.values() for name in dist])
```
maxhully: The `set()` constructor can take the plain generator expression instead of the list:
```suggestion
        assgn_names = set(name for dist in self.assignment.parts.values() for name in dist)
```

gerrychain/partition/partition.py
```diff
@@ -41,6 +41,10 @@ def _first_time(self, graph, assignment, updaters):
 
         self.assignment = get_assignment(assignment, graph)
 
+         if not self.validate_assignment():
```
maxhully: Looks like there's an extra space before this `if` that's causing the failing tests. You can fix it manually or outsource it to a formatter like [black](https://black.readthedocs.io/en/stable/).

zschutzman: edited manually

zschutzman:
I've committed the changes and fixed the formatting problem.  If `NameError` is okay, then I think it's good to go?

PR #285, opened 2019-03-14
Missed case in assignment checking

zschutzman:
The stuff I wrote yesterday doesn't handle the case where all of the unit names are correct, but the assignment has some unit in more than one district. This checks for that problem as well.

### DIFF
--- a/gerrychain/partition/partition.py
+++ b/gerrychain/partition/partition.py
@@ -56,6 +56,9 @@ class Partition:
 
     def validate_assignment(self):
         node_names = set(self.graph.nodes)
+        if len(node_names) != sum(len(dist) for dist in self.assignment.parts.values()):
+            return False
+
         assgn_names = set(name for dist in self.assignment.parts.values() for name in dist)
         return node_names == assgn_names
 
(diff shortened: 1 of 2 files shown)


maxhully:
We should add tests covering these behaviors too.

zschutzman:
I don't think that this last case is even reachable in the current setup, since the constructor inverts a dictionary of the form {vertex : assn}, and a vertex can't appear as a key more than once, so it will never be in two different sets in the reversed dictionary.  It may become an issue as more kinds of i/o are allowed and we need to worry about a user accidentally modifying the second column of a csv, for example.

I can write a test function to try the first two cases (not every vertex has an assignment; dictionary names are different from vertex names).

maxhully:
Good point! Then we should probably rearrange things so that case raises an exception.

zschutzman:
I wrote three test functions. The first two check that doing something bad with standard methods causes an error.  The third verifies that the `validate_assignment()` method is catching the case where some vertex belongs to more than one district, even though this is not reachable in the typical workflow.

PR #301, opened 2019-05-08
Slow version of the actual reversible flip

drdeford:
Temporary version of the pick a (node/part) pair version of the flip until we decide what we want to do about the updater version. 

Also, a spectral version of recom (meant to put it in a separate pull request but got my branches swapped). Currently just provides a separate function since we hadn't settled on the right way to have all the recoms swap out the bipartioning methods given that they need different input parameters.


maxhully:
Thanks Daryl!

PR #310, opened 2019-06-25
Make ElectionResults.percents return a dictionary

rwbogl:
This makes it easier to use dataframes for election percents, since `from_dict` is now an option.


rwbogl:
Some thoughts about this that came up today:

1. This breaks a lot of tests which relied on the tuple behavior.
2. This breaks the ReCom documentation [here](https://gerrychain.readthedocs.io/en/latest/user/recom.html#running-the-chain).

Is this worth doing if it substantially modifies behavior?

Issue #312, opened 2019-06-28
pop_target not updated in recursive_tree_part()


pjrule:
I've been using `recursive_tree_part()` to split North Dakota's Census VTDs into 10 districts with an epsilon of 15%. With a population target of 67259.1, I've seen many runs where the first 9 districts (and more generally, the first _n_-1 districts) are allocated with populations below the target, resulting in the leftover voters being stuffed into the _n_-th district. Here are the populations from an example run:
```
   {0: 62966,
    1: 73140,
    2: 58006,
    3: 57653,
    4: 68164,
    5: 68149,
    6: 63184,
    7: 63916,
    8: 57318,
    9: 100095}
```

This seems to partially relate to the fact that `pop_target` is not dynamically updated as the graph is split. The code notes this:
https://github.com/mggg/GerryChain/blob/cf09999a1281b457e07d901b4b082cb189bb6936/gerrychain/tree.py#L217

@maxhully I'd be happy to submit a PR to fix this if we think it's a notable issue. Let's chat. :)

maxhully:
Closed by #317 !

PR #317, opened 2019-07-01
Add population debt to recursive_tree_part()

pjrule:
The `recursive_tree_part()` function attempts to partition a population-labeled labeled graph with _n_ people into _m_ subgraphs (districts) with some population tolerance ϵ. The target population for each subgraph is _T = n/m_.  Over many runs, I've observed a consistent behavior where the first _m-1_ subgraphs all end up with populations below the target; accordingly, the _m_-th subgraph wildly overshoots the target population. It is also possible for the reverse effect to occur, where the _m_-th subgraph _undershoots_ the target population. These issues occur because the current version of the function makes no attempt to keep track of accumulated error (see issue #312). The maximum error is ± (_m-1_)ϵ—that is, maximum error is linear with the number of districts—so this issue is particularly evident in cases where many subgraphs exist, such as state legislatures and city councils.

This PR provides a version of `recursive_tree_part()` that imposes a stricter population tolerance regime. The function keeps track of a "population debt" to guarantee that all subgraph populations are within ±ϵ of the target population.

More specifically, we define district _i_'s debt (_dᵢ_) as the amount by which district _i_'s population (_pᵢ_) exceeds the target population _T_. (Thus, _dᵢ_ is negative if _pᵢ_ comes in under the population tolerance). At each partition step, we set the next district's bounds such that the last district's debt must be paid off. This guarantees that error exceeding ±ϵ cannot accumulate.

As a concrete example, consider a graph with 500 people, which we'd like to split into 5 subgraphs with a 5% population tolerance. Here's how the debt might accumulate:

|Subgraph|Bounds|Allocated pop.|Cumulative pop.|Debt|
|--------|------|--------------|---------------|----|
|1|[95, 105] |105|105|5|
|2|[95, 100] |100|205|5|
|3|[95, 100] |95|300|0|
|4|[95, 105] |97|397|-3|
|5|[103, 103]|103|500|0|

To test these changes, I used `recursive_tree_part()` to build 50-ward Chicago city council districting plans using [MGGG's shapefile](https://github.com/mggg-states/IL-shapefiles/tree/master/Chicago). With a negligible effect on failure rate, the modified algorithm ensures that all district populations are within the population tolerance:

chicago_ward_pops

(*Edit*: changed the bounds in the first row of the table from [195,105] to [95,105].)

### DIFF
--- a/gerrychain/tree.py
+++ b/gerrychain/tree.py
@@ -202,20 +202,31 @@ def recursive_tree_part(
     """
     flips = {}
     remaining_nodes = set(graph.nodes)
+    # We keep a running tally of deviation from ``epsilon`` at each partition
+    # and use it to tighten the population constraints on a per-partition
+    # basis such that every partition, including the last partition, has a
+    # population within +/-``epsilon`` of the target population.
+    # For instance, if district n's population exceeds the target by 2%
+    # with a +/-2% epsilon, then district n+1's population should be between
+    # 98% of the target population and the target population.
+    debt = 0
 
     for part in parts[:-1]:
+        min_pop = max(pop_target * (1 - epsilon), pop_target * (1 - epsilon) - debt)
+        max_pop = min(pop_target * (1 + epsilon), pop_target * (1 + epsilon) - debt)
         nodes = method(
             graph.subgraph(remaining_nodes),
             pop_col=pop_col,
-            pop_target=pop_target,
-            epsilon=epsilon,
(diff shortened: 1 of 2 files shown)


rwbogl:
I haven't looked too closely, but the results here seem awesome. Very impressive, glad you looked into this!

maxhully:
Looks good! Merging in now.

Issue #322, opened 2019-08-08
Graph.from_geodataframe() doesn't add column data


drdeford:
Unlike .from_file there isn't a separate call to add data when you use from_geodataframe directly. Indeed, .from_file calls add data immediately after building the graph object with from_geodataframe. 

Is there a good reason to not have from_geodataframe take the cols_to_add argument and do everything at once?

zschutzman:
I can't think of a good reason, especially because `from_file` calls `from_geodataframe`.

This seems like a reasonable thing to implement if someone wants to do the pull request, or I can do it when I get back from China.

drdeford:
I'll take care of it.

drdeford:
Solved by @zschutzman  in PR #324

zschutzman:
Fixed in #324.

PR #324, opened 2019-08-29
`from_geodataframe` calls `add_data()`

zschutzman:
This PR updates the `from_geodataframe` constructor to call `add_data()` in the same way that `from_file` does.  

Addresses issue #322


drdeford:
Happy to merge now!

PR #331, opened 2020-02-04
TOTPOP column from MGGG-states

drdeford:


drdeford:
More doc fixes to match up with changes in the underlying mggg-states data.

PR #337, opened 2020-09-01
Fix errors in unit tests

ptrks:
Addressed issues related to the 3 failing and 1 error when running unit tests.

The first major group of errors had to do with accessing the deprecated `node` property, which is now `nodes`.  The second error related to a unit test attempting to delete the `crs` property of a `geodataframe` object. This was solved by setting the property to `None` which should simulate the same failure behavior. 

Please let me know if you have any questions or recommendations on modifications to this PR.


bsuwal:
A bit late, but thanks so much @ptrks ! Merged!

PR #343, opened 2021-01-13
use Conda for CircleCI tests

bsuwal:


bsuwal:
Some tests were failing -- for whatever reason Shapely 1.7.1 isn't happy with GeoPandas and errors out. I simply re-install Shapely v1.6.4 and the tests pass. En route to this I also changed the CircleCI to use Conda instead of Pip because installing from Pip wouldn't fix the tests for some reason. I am not tooo familiar with Python package versioning to know why exactly this is -- but I asked a  [stackoverflow question](https://stackoverflow.com/questions/65676484/cannot-make-geodataframe-from-shapely-polygons-notimplementederror-a-polygon-d) and didn't get anywhere.

PR #358, opened 2021-06-07
Fix reversible ReCom

pjrule:
We've discovered a minor issue with how probabilities are computed in the last rejection step of our reversible ReCom implementation. This PR brings the code in sync with our current draft of the paper.

### DIFF
--- a/gerrychain/proposals/tree_proposals.py
+++ b/gerrychain/proposals/tree_proposals.py
@@ -5,7 +5,6 @@ from ..tree import (
     recursive_tree_part, bipartition_tree, bipartition_tree_random,
     _bipartition_tree_random_all, uniform_spanning_tree,
     find_balanced_edge_cuts_memoization,
-    BalanceError
 )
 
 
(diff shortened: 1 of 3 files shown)


gabeschoenbach (commented):
Trying to see if I understand what's going on here, let me know if I've got it right:
- In `tree.py`, we add the `_bipartition_tree_random_all()` internal function, which essentially draws out the computation of all possible balanced cuts that used to be in `bipartition_tree_random()`. This new internal function returns all possible balanced cuts. Now, `bipartition_tree_random()` works exactly the same way, but is virtually empty — it just calls the internal function and then randomly chooses one balanced cut edge to return, if any exist.
- The changes in `tree_proposals.py` are actually what change the output of the chain. We can now use the output of our internal `_bipartition_tree_random_all()` function to learn exactly how many possible cuts there are, and then condition our acceptance probability in direct proportion to that number. We still randomly choose one of those cut edges to actually make our bipartition.

On a high level, this code looks good, though I'm not totally sure how to test it. One thing that still isn't clear to me is the code around 122-126 in `tree_proposals.py`: what is the math behind the upper bound `M` and its relationship to the seam length and the number of possible balance edges?

pjrule:
@gabeschoenbach Yes, that's right! Also, the math behind the seam length is in our (soon-to-be-posted) RevReCom paper—stay tuned... 👀

InnovativeInventor (approved):
Looks great! I'll trust you on the math/correctness -- can't wait to see your paper!

PR #354, opened 2021-03-16
add get_num_spanning_trees function

blumt:

PR #359, opened 2021-07-06
Move spanning tree counter to updaters

pjrule:
We recently introduced a function to count spanning trees (#354). For consistency with the rest of the GerryChain codebase, this function should be an updater, rather than a method in the `Partition` class. 

This PR moves `get_num_spanning_trees` to `gerrychain.updaters` (as `num_spanning_trees`). This new updater returns spanning tree counts for _all_ districts, rather than a single district.

### DIFF
--- a/gerrychain/partition/partition.py
+++ b/gerrychain/partition/partition.py
@@ -113,7 +113,6 @@ class Partition:
 
     def __getattr__(self, key):
         return self[key]
-
     def keys(self):
         return self.updaters.keys()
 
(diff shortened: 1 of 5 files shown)


gabeschoenbach (commented):
Looks good! Only one question: I could imagine making the `round()` operation happen inside of `num_spanning_trees` rather than outside of it. If we were to put it inside:

Pros:
- Conform to user expectations of having the output of the updater be integer-valued (though still of type `float` I believe). 
- Seems slightly cleaner, so users don't need to remember to `round()` the output of this updater if they want it to be so.

Cons:
- This might make it easy for the user to _forget_ that this method of computing the number of spanning trees has some potential for error (floating point error? I don't know exactly how to describe this). We might want to keep users aware of these difficulties.

pjrule:
@gabeschoenbach Thanks for the review! IMO, it's best to keep the spanning tree counts as `float`s to avoid giving a sense of false precision—floating-point error starts to actually matter here for large graphs because the spanning tree counts are so enormous. (If you're curious, I have a notebook that compares exact spanning tree counts derived using `sympy` with the very close numerical approximation we get here.) That said, we may want to update the documentation to reflect this.

@InnovativeInventor @jenni-niels Thoughts?

InnovativeInventor (approved):
Looks good to me -- this is a little short on correctness tests, but that's perhaps out of scope of this PR.

PR #361, opened 2021-07-29
Add Common Optimization Methods

jenni-niels:
This PR adds common optimization methods to the gerrychain codebase.
The `SingleMetricOptimizer` class represents the class of optimization problems over a single plan metric and currently implements short bursts, a few variants, and tilted runs, with more to come.

The `Gingleator` class is a subclass of `SingleMetricOptimizer` and can be used to search for plans with increased numbers of Gingles' districts.


apizzimenti:
I have no comments (for now) other than `Gingleator` is a great class name

gabeschoenbach (dismissed):
I think this is a great jumping-off point for us to bring short bursts into harmony with guided acceptance functions!

I like the flexibility of thresholding our searches — I'm going to keep saying "majority" (i.e. `threshold = 0.50`) for simplicity but this should all work if the threshold is set to something else). But I think we should broaden the `gingleator` initialization to be able to search for majority-minority **or** majority-party districts. This flexibility is sketched out [here](https://github.com/gabeschoenbach/One-Click-Chains/blob/main/score_functions.py) and [here](https://github.com/gabeschoenbach/One-Click-Chains/blob/main/acceptance_functions.py) in my One-Click-Chains repo, though admittedly not in an "object-oriented" format. While broadening to partisan stuff makes the code a little less straightforward, I think we could gain that back by altogether dropping the `minority_perc_col` and just expect to be passed either a Tally updater name for the demographic group (along with a Tally updater for the total population) or an ElectionResults updater name that we can use `.percents(party)` to query for the party percents.

Conceptually, one thing all of the score functions in the `gingleator` class have in common is that they look built to be used as simple comparators as you traverse the chain, i.e. accept the child if there is an improvement, otherwise accept with some fixed probability _p_. I think we should build out the functionality to make _p_ dynamically change depending on how much worse the proposal is than its parent. In `gingleator` terms, this means a helper function like my `get_majdistricts_info()` function that returns the number of majority-{group} (demographic group or party) districts, and the percentages of a) the smallest district above the threshold and b) the largest district below the threshold. Then, in `SingleMetricOptimizer` we can build acceptance functions that use those helpers to cleverly accept with a variable _p_. As I see it, this would be an extension/improvement to your `tilted_short_bursts()`.

This is really exciting — I think that if we build this right we can be really flexible in how we search through the metagraph, and I would love to run experiments to see if layering all of these tricks together gives us more of a leg up (variable length short bursts with a custom acceptance function that rejects worse plans proportional how bad they are?? could be huge)...

Small bug, I think:
```python
        if minority_perc_col is None:
            perc_up = {min_perc_column_name:
                            lambda part: {k: part[minority_pop_col][k] / part[total_pop_col][k]
                                          for k in part.parts.keys()}}
            initial_state.updaters.update(perc_up)

        score = partial(score_function, minority_perc_col=minority_perc_col, threshold=threshold)

        super().__init__(proposal, constraints, initial_state, score, minmax="max",
                         tracking_funct=tracking_funct)

    """
    Score Functions
    """

    @classmethod
    def num_opportunity_dists(cls, part, minority_perc_col, threshold):
        """
        Given a partition, returns the number of opportunity districts.
        :param `part`: Partition to score.
        :param `minority_perc_col`: Which updater is a mapping of district ids to the fraction of
            minority popultion within that district.
        :param `threshold`: Beyond which fraction to consider something a "Gingles" 
            (or opportunity) district.
        :rtype int
        """
        dist_percs = part[minority_perc_col].values()
        return sum(list(map(lambda v: v >= threshold, dist_percs)))
```

...if `minority_perc_col` is `None` then the updater that maps district IDs to the fraction of minority population will be called `minority_perc_column_name`. But the score functions all seem to call `partition[minority_perc_col]` which seems like it would return an error in this case.

gabeschoenbach:
This looks really great! Just to check my understanding — if we called a simulated annealing run with a `beta_function` as something like:
```python
gingles.hot_cold_cycle_beta_function_factory(0,1000)
```
(in other words only ever cold), would this be equivalent to a tilted run that always accepts better partitions, and accepts worse partitions with a _dynamic_ probability `p` that depends on the `beta_magnitude`?

I made some small changes in docstrings, mostly just fixing some typos. I also want to flag a couple spots I think the documentation is unclear — might just be me, so would love to get other folks' input as well...

**SingleMetricOptimizer** `optimizer.py, lines 12-25`
I would change `In instance of this class encapsulates the dualgraph and updaters via the initial partition` to `This class includes the initial partition (which gives access to the underlying dual graph and updaters)...`. I'm a little confused by `Note that these are reset every time an optimization run is invoked and do not persist`, but I'm not sure whether/how to reword that.

**hot_cold_cycle_beta_function_factory** `optimizer.py, lines 137-144`
I wonder if we can think of a more concise name for this function? And maybe add one more sentence of explanation as to its use, although I suppose this is pretty clear if you read the args on lines 140-141...

The `Optimization` notebook looks great! I made a small change to increase the size of the traceplots, so it's easier to see the differences in the different chains. I think it could be useful to add a little bit more documentation to the cells of the notebook, so someone could understand how the various functions work without going to the documentation. This is something I could add if you don't have the bandwidth!

gabeschoenbach:
Definitely agree it would be good to show `SingleMetricOptimization` for other things — aggregate proportionality would make sense. I also was thinking it would be cool to compare/contrast all these different methods, but I think that would be best in a different file, maybe not necessarily an intro notebook. If we do do that comparison, one thing I'd love to try is just optimizing for cut edges, since for large graphs its a pretty granular metric and it would be easy to see change over time.

gabeschoenbach:
Looks good! I just updated some stuff in the `Optimization` notebook so the annealing calls work with the new `jumpcycle` function.

PR #362, opened 2021-08-17
Region-aware ReCom

gabeschoenbach:


gabeschoenbach:
This PR was largely based on the ideas and work of @amybecker and @drdeford, on how to update the ReCom proposal function to be aware of county boundaries. @phorva01 and I extended these ideas to allow ReCom to be aware of multiple regions at once — say, counties and municipalities. 

At the highest level, we add an optional parameter called `region_weights` to the `recom()` function in `tree.py`. If included, it should be a list of tuples, like `[("COUNTYFP20", 1), ("TOWN", 1)]`. This would tell ReCom to try to create districts whose boundaries fall along county and town boundary lines, to the extent possible while constrained by population balance concerns. 

This works in two ways:
1. In `tree.py`'s `random_spanning_tree()`, we randomly assign edge weights to our graph in order to draw a minimum spanning tree and look for an edge to cut the graph into two population-balanced parts. If `region_weights` is not `None`, we will add extra weight to edges that cross our regions, therefore encouraging our spanning tree to be drawn mostly inside our regions, with as few edges crossing region boundaries as possible. Regardless of where the balance edge is, this will help keep regions whole. See below for an example of (a) a minimum spanning tree drawn randomly, and (b) a minimum spanning tree drawn with counties in mind
mst_edges_with_counties
aware_edges_with_counties
2. In `tree.py`'s `find_balanced_edge_cuts_memoization()`, we search for an edge in the minimum spanning tree to cut the graph into two equal-sized parts. Ideally, this edge is one of the few in the tree that _does_ cross from one region to another, since this would neatly split the districts along a region boundary. We first look to see if there exists such an edge, but if not we simply look for any balance edge. If we care about multiple regions, we check to see if there is an edge that crosses as many of our regions as possible.

To sum up, the optional `region_weights` parameter in the `recom()` proposal function allows the user to specify an arbitrary number of regions for GerryChain to attempt to keep intact. The weights corresponding to each region will be used in the minimum spanning tree step. If one is more interested in keeping counties intact than towns, they could make the weight for counties larger than that for towns. This would mean ReCom would first look for balanced edges that span different counties in step 2. However, our testing has shown that setting weights of 1 for every region does quite well in optimizing intactness for all region-types simultaneously.

To test out this code, try working through the Jupyter Notebook here.
[testing_region_aware.ipynb.zip](https://github.com/mggg/GerryChain/files/7012012/testing_region_aware.ipynb.zip)

gerrychain/tree.py
```diff
@@ -151,11 +159,37 @@ def part_nodes(start):
...
-        elif abs((total_pop - tree_pop) - h.ideal_pop) <= h.ideal_pop * h.epsilon:
-            cuts.append(Cut(edge=(node, pred[node]),
-                            subset=set(h.graph.nodes) - part_nodes(node)))
+        is_balanced_A = abs(tree_pop - h.ideal_pop) <= h.ideal_pop * h.epsilon
+        is_balanced_B = abs((total_pop - tree_pop) - h.ideal_pop) <= h.ideal_pop * h.epsilon
+
+        parent = pred[node]
+        if region_weights is not None:
+            region_cols = [tup[0] for tup in region_weights]
+            # score function to prefer dividing higher ranked regions. three regions -> [4, 2, 1]
```
pjrule, line 168: It would be nice to flesh out the example here a little more rather than using shorthand (e.g. `three regions -> [4, 2, 1]`)

gerrychain/tree.py
```diff
@@ -151,11 +159,37 @@ def part_nodes(start):
...
+            # an edge that splits our first- and last-ranked regions would get a score of 4 + 1 = 5
+            # if node corresponds to a balance cut and our split score is as good or better than our
+            # best previously seen split score, add this cut to our list of potential balance cuts
+            node_split_score = 0
+            for i, region_col in enumerate(region_cols):
+                if h.graph.nodes[parent][region_col] != h.graph.nodes[node][region_col]:
+                    node_split_score += 2 ** (len(region_cols) - i - 1)
```
pjrule, line 175: If we're reading this right, the weights contributed by the regions depends on the _ordering_ of `region_weights`, not on the right values of the 2-tuples in `region_weights`. This is confusing because it's undocumented—as a user, there's no _a priori_ reason for me to believe  `region_weights=[("COUNTYFP20", 1), ("TOWN", 1)]` will yield distinct behavior from `region_weights=[("TOWN", 1), ("COUNTYFP20", 1)]`.

gerrychain/tree.py
```diff
@@ -164,16 +198,20 @@ def bipartition_tree(
...
     pop_target,
     epsilon,
+    region_weights=None,
     node_repeats=1,
     spanning_tree=None,
     spanning_tree_fn=random_spanning_tree,
     balance_edge_fn=find_balanced_edge_cuts_memoization,
-    choice=random.choice
+    choice=random.choice,
+    attempts_before_giveup=100,
```
pjrule, line 207: Simplify to `attempts`?

gerrychain/tree.py
```diff
@@ -151,11 +159,37 @@ def part_nodes(start):
...
+            # score function to prefer dividing higher ranked regions. three regions -> [4, 2, 1]
+            # an edge that splits our first- and last-ranked regions would get a score of 4 + 1 = 5
+            # if node corresponds to a balance cut and our split score is as good or better than our
+            # best previously seen split score, add this cut to our list of potential balance cuts
+            node_split_score = 0
+            for i, region_col in enumerate(region_cols):
+                if h.graph.nodes[parent][region_col] != h.graph.nodes[node][region_col]:
+                    node_split_score += 2 ** (len(region_cols) - i - 1)
+
+            if node_split_score > best_split_score and (is_balanced_A or is_balanced_B):
```
pjrule, line 177: Oddly enough, this might be more clear if it was nested differently to avoid handling the symmetry twice, e.g.
```python
if node_split_score >= best_split_score and (is_balanced_A or is_balanced_B):
  if node_split_score > best_split_score:  # special case: strictly better
    best_split_score = node_split_score
    cuts = []
  part_subset = part_nodes(node) if is_balanced_A \
                    else set(h.graph.nodes) - part_nodes(node)
  cuts.append(Cut(edge=(node, pred[node]), subset=part_subset))
```

pjrule (changes requested):
(From review by me and @apizzimenti)

Overall, this PR looks good! However, it modifies several core functions (notably, `recom` and  `recursive_tree_part`) that many of our users interact with directly. If users are calling `recom` with positional arguments instead of keyword arguments, for instance, this could introduce a breaking change, as `region_weights` is inserted in the middle of the existing argument order. As a general principle, we think it is best to separate existing core functionality from these new features. A way to refactor this code would be:

* Preserve the existing contracts of `random_spanning_tree` and `recom`.
* Create new functions for region-aware stuff:
*   `random_region_aware_spanning_tree` should contain region-aware weighting logic and take a `region_weights` parameter. 
* Similarly, `region_aware_recom` should take a `region_weights` parameter. 

This way, we have parallel structure with reversible ReCom.

Additional things we might consider tweaking/things to consider:
* Why is the representation of `region_weights` a list of 2-tuples? Would it be more intuitive to use a dictionary mapping between columns and weights? (Note that such a representation can be converted to the current representation with `list(region_weights.items())` or the like.)

InnovativeInventor:
@gabeschoenbach @phorva01 I just fixed the merge conflict. @pjrule @apizzimenti do you think your concerns have been suitable addressed? What's the status on this PR?

gabeschoenbach:
I updated the `region_weights` to a dictionary now, which I agree is a much friendlier input type! The last thing I'm struggling with his how to correctly prioritize balance edges in the cutting step. I feel good about our power-of-two idea whenever there are no tied penalty weights, but as noted above it doesn't work in cases where (say) we have two regions each with penalties of 1. What our code does is first look for balance edges that cut across both region types, but then it prioritizes balance edges that only cut across the first listed region type. Ideally we'd want it to look for balance edges that cut across either region type and choose uniformly from that set. But I'm not sure the nicest way to get that sort of functionality! Any thoughts?

gabeschoenbach:
@jenni-niels I just added a commit that keeps track of the edge penalties and only adds it to the `cuts` list (from which one balance edge will be selected at random) if it has the highest penalty of all the balance edges in the tree. I think this should work the same way as you were describing your .NET implementation works, but please take a look and let me know!

peterrrock2:
Thank you for all of your work on this! We decided to add the option to do region aware as a core part of the functionality of the `recom` function in GerryChain in v0.3.0. The implementation was slightly different than what is outlined here, but if you would like me to add you to the release notes for credit on the work that you have done, please let me know!

gabeschoenbach:
> Thank you for all of your work on this! We decided to add the option to do region aware as a core part of the functionality of the `recom` function in GerryChain in v0.3.0. The implementation was slightly different than what is outlined here, but if you would like me to add you to the release notes for credit on the work that you have done, please let me know!

Great to hear, glad the functionality has been incorporated! Yes, I'd love a shoutout in the release notes :)

peterrrock2:
> > Thank you for all of your work on this! We decided to add the option to do region aware as a core part of the functionality of the `recom` function in GerryChain in v0.3.0. The implementation was slightly different than what is outlined here, but if you would like me to add you to the release notes for credit on the work that you have done, please let me know!
> 
> Great to hear, glad the functionality has been incorporated! Yes, I'd love a shoutout in the release notes :)

Done! You are now in the release notes for version 0.3.0!

PR #363, opened 2021-08-30
Use __slots__ in Partition

pjrule:
_(I'm not sure if this is worth it yet, but figured I'd throw up a PR and we can do some benchmarks.)_ I was just reminded that using `__slots__` is known to improve memory usage (and sometimes performance) of Python classes [(see here)](https://book.pythontips.com/en/latest/__slots__magic.html), at least when the classes are small. I suspect our memory usage is dominated by large assignment dictionaries, but maybe this is a step in the right direction. `__slots__` also protects against abuse of the `Partition` class by preventing the use of additional fields, though this may break some people's code.

### DIFF
--- a/gerrychain/partition/partition.py
+++ b/gerrychain/partition/partition.py
@@ -16,8 +16,18 @@ class Partition:
     :ivar dict parts: Maps district IDs to the set of nodes in that district.
     :ivar dict subgraphs: Maps district IDs to the induced subgraph of that district.
     """
-
     default_updaters = {"cut_edges": cut_edges}
+    __slots__ = (
+        'graph',
+        'subgraphs',
+        'assignment',
+        'updaters',
+        'parent',
+        'flips',
+        'flows',
+        'edge_flows',
+        '_cache'
+    )
 
     def __init__(
         self, graph=None, assignment=None, updaters=None, parent=None, flips=None
--- a/tests/constraints/test_validity.py
+++ b/tests/constraints/test_validity.py
@@ -14,27 +14,35 @@ from gerrychain.partition.partition import get_assignment
 
 
 @pytest.fixture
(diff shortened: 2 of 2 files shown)


InnovativeInventor:
@pjrule I'm going to give this a rebase and run some performance benchmarks. If we see a sizable improvement in replay speeds, I think it'd be worth merging in.

InnovativeInventor:
With the benchmarking utils I wrote a while back to look for performance regressions, I can confirm that this simple change results in a *massive* speedup in replay times. I estimate this gives a ~2-3x speedup in replay performance.

Replay performance without `__slots__`:
```
10 loops, best of 5: 1.42 sec per loop
```

Same chain replay performance with `__slots__`:
```
10 loops, best of 5: 610 msec per loop
```

We should totally merge this in!

InnovativeInventor (approved):
Excellent speedup in replay times!

InnovativeInventor:
Note that GerryChain is still compatible with Python 3.7, so we shouldn't remove it from the list of supported Python versions (I've been able to get GerryChain to work with Python 3.7 with some manual dependency wrangling).

InnovativeInventor:
> With the benchmarking utils I wrote a while back to look for performance regressions, I can confirm that this simple change results in a massive speedup in replay times. I estimate this gives a ~2-3x speedup in replay performance.

So . . . turns out this is not true. I was benchmarking off an old commit in `main` (before #367 was merged in), so the amended benchmarks on 10k replays are as follows:

- 194 it/s (Parker's slots)
- 189 it/s (main branch)

It's still worth it, though (every little bit counts).

InnovativeInventor:
Note that once we merge this in, our next release should be a major one (we should follow semver).

PR #372, opened 2022-01-02
Performance enhancements

InnovativeInventor:
This is a work-in-progress, draft PR with some performance enhancements. In particular,
- `__slots__` is enabled for a few frequently-created classes (`Assignment` and `SubgraphView`), resulting in minor performance improvements
- caching for iterating over the neighbors of all flips is implemented to improve flow and cut edge calculation runtimes
- caching for iterating over the neighbors of a node in a graph is implemented (since `networkx`'s neighbors function is quite expensive and a partition's graph will always remain static)
- `Graph.nodes` and `Graph.edges` is cached

Combined, these speedups improve *both* standard GerryChain run performance and replay performance. In particular, the time it takes to run 100 steps on PA goes from 29.8 seconds to 26 seconds (~13% decrease) and the time spent on each loop when replaying goes from 610 ms (based on #363) to 469 ms (~23% decrease).

Some caveats:
- I used `functools.cache`, which is a Python 3.9 feature. We should switch to `functools.lru_cache` or something with better compatibility with older versions of Python before we merge this in.
- For caching reasons, GerryChain Partitions will no longer accept a `networkx` graph (to convert a `networkx` graph to a `gerrychain.Graph` object, a new method `Graph.from_networkx` is introduced in this PR). I had to rewrite the tests accordingly.

### DIFF
--- a/gerrychain/constraints/contiguity.py
+++ b/gerrychain/constraints/contiguity.py
@@ -78,7 +78,7 @@ def single_flip_contiguous(partition):
         """Compute the district edge weight, which is 1 if the nodes have the same
         assignment, and infinity otherwise.
         """
-        if assignment[start_node] != assignment[end_node]:
+        if assignment.mapping[start_node] != assignment.mapping[end_node]:
             # Fun fact: networkx actually refuses to take edges with None
             # weight.
             return True
@@ -86,12 +86,12 @@ def single_flip_contiguous(partition):
         return False
 
     for changed_node in flips:
-        old_assignment = partition.parent.assignment[changed_node]
+        old_assignment = partition.parent.assignment.mapping[changed_node]
 
         old_neighbors = [
             node
             for node in graph.neighbors(changed_node)
-            if assignment[node] == old_assignment
+            if assignment.mapping[node] == old_assignment
         ]
 
         if not old_neighbors:
@@ -130,7 +130,7 @@ def affected_parts(partition):
     affected = set()
     for node, part in flips.items():
         affected.add(part)
-        affected.add(parent.assignment[node])
+        affected.add(parent.assignment.mapping[node])
 
     return affected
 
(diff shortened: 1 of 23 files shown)


InnovativeInventor:
Note that the behavior of `Graph.nodes` and `Graph.edges` will not be backwards-compatible. It may be better to rename the cached versions to `Graph._cached_nodes` and `Graph._cached_edges`.

InnovativeInventor:
Some more benchmark stats with 11906d0 (replay perf on 1000-step chains, PA congressional):
- this PR (#372): 220 it/s
- Parker's PR (#363): 170 it/s
- main branch: 72 it/s

The `FrozenGraph` class also opens the door to using faster read-only data structures (generated when the class is instantiated) behind the scenes without breaking backwards-compatibility.

InnovativeInventor:
Updated stats (wrote some better benchmarks):
```
- 247 it/s (my perf tuning + Parker's slots)
- 188 it/s (Parker's slots)
- 82 it/s (main branch)
```

InnovativeInventor:
Note: I will break this up into multiple PRs, so it'll be easier to review.

InnovativeInventor:
I realized that I was profiling #363 incorrectly -- the performance is now as follows:
- 353 it/s (my perf tuning + Parker's slots)
- 194 it/s (Parker's slots)
- 189 it/s (main branch)

My bad -- looks like `__slots__` doesn't help that much after all.

InnovativeInventor:
Memory leak!: https://github.com/mggg/GerryChain/blob/c55f38260b196cc1a20c1d0c28b6eb1dbbb0b35c/gerrychain/partition/partition.py#L121

InnovativeInventor:
Reminder to self: `Partition.crosses_parts` caching increases performance (but the previous impl introduced a memory leak).

InnovativeInventor:
Closing due to the successful merges of all the broken out PRs. The retworkx PR is still a work-in-progress, but will be ready soon.

PR #373, opened 2022-01-06
Add neighborhood flips caching

InnovativeInventor:
This is a PR broken out from #372 to ease the code review process.

In particular, this patch is intended to improve the performance of calculating `cut_edges` by introducing a `neighbor_flips` function to cache the set of neighbors of all flipped nodes. This is used to calculate the cut edges of a plan as the set of all flipped nodes is a superset of the cut edges. 

Since this is looped over a few times throughout the lifetime of a Partition object, caching improves replay performance considerably.

### DIFF
--- a/gerrychain/updaters/flows.py
+++ b/gerrychain/updaters/flows.py
@@ -2,6 +2,16 @@ import collections
 import functools
 
 
+@functools.lru_cache(maxsize=2)
+def neighbor_flips(partition):
+    """The neighbors of the flips in a partition"""
+    return {
+        tuple(sorted((node, neighbor)))
+        for node in partition.flips
+        for neighbor in partition.graph.neighbors(node)
+    }
+
+
 def create_flow():
     return {'in': set(), 'out': set()}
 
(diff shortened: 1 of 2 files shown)


gerrychain/updaters/flows.py
```diff
@@ -2,6 +2,16 @@
 import functools
 
 
+@functools.lru_cache(maxsize=2)
+def neighbor_flips(partition):
+    """The neighbors of the flips in a partition"""
+    return {
+        tuple(sorted((node, neighbor)))
```
apizzimenti, line 9: I understand that the reason vertices are sorted is to prevent duplicate edges – is there a better way to do this? There may not be – and it's probable that it wouldn't make much of a difference anyway – but it's something to think about.

apizzimenti (approved):

gerrychain/updaters/flows.py
InnovativeInventor: I don't think there is a faster way to remove duplicate edges (I tried some other methods, but this seemed to work the best and guarantee equivalence with the previous version of the code).

PR #374, opened 2022-01-06
More __slots__

InnovativeInventor:
This is a PR broken out from #372 to ease the code review process.

In particular, this patch adds the `__slots__` method to a few classes that are created often, which improves memory and runtime performance (marginally but measurably). In theory, this is a breaking change, as users will no longer be able to add arbitrary attributes to the `Assignment` and `SubgraphView` class. However, I don't think this is something that people do (it would be quite strange, no?).

This PR inspired by @pjrule #363. See that PR for more details on the reasoning.


apizzimenti (approved):
These changes match the design patterns from #363 and improve performance. Seems like a good addition to the codebase.

PR #378, opened 2022-01-06
Add seed and freeze test to loudly detect when breaking changes are made

InnovativeInventor:
Some time ago, I flagged a few historical commits that unintentionally produce different chain outputs (with the same seed, etc). This is concerning (since it could mean that correctness issues were introduced in those commits). This PR adds a "seed and freeze" test to loudly detect such changes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ jobs:
 
             # run tests
             echo "backend: Agg" > "matplotlibrc"
-            pytest -v --cov=gerrychain --junitxml=test-reports/junit.xml tests
+            pytest -v --runslow --cov=gerrychain --junitxml=test-reports/junit.xml tests
             codecov
           environment:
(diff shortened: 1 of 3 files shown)


InnovativeInventor:
If anyone's watching this PR and is curious about the list of commits that changed GerryChain outputs (and were flagged for careful correctness review), here it is:

456eddca569385635f9f75934449b95c8e87b305
8e8a73fd1116dac67ce19971bdac1e468af31a69
e2ec301b7de9fc1c74ce6cc0511b4b55eae59a10
77b6ede812056d83b29ad5f0c5a03e4a0734e0dd
b7f0289ac5743b9d834f4feb0619a37c882b04e7
dc2896f0360e1aac3f8d35b2482af994f8b611f2
01b8801877498f87d1b9968415bc7c2439f448c0

Note that a fair number of these commits *appear* to be correct. However, they're totally worth a second look.

PR #382, opened 2022-01-19
Replace references to assignment[node] to assignment.mapping[node] to reduce __getitem__ overhead

InnovativeInventor:

### DIFF
--- a/gerrychain/constraints/contiguity.py
+++ b/gerrychain/constraints/contiguity.py
@@ -78,7 +78,7 @@ def single_flip_contiguous(partition):
         """Compute the district edge weight, which is 1 if the nodes have the same
         assignment, and infinity otherwise.
         """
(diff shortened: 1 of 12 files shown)


gabeschoenbach (approved):
This looks all right to me, but I have a question — if `assignment` is a dictionary, where does the `.mapping` attribute come from?

InnovativeInventor:
@gabeschoenbach it comes from the `Assignment` class. In particular, it is instantiated here: https://github.com/mggg/GerryChain/blob/8acd44c694381e3d6c4e317150095e2e4619caad/gerrychain/partition/assignment.py#L35-L41

and updated here:
https://github.com/mggg/GerryChain/blob/8acd44c694381e3d6c4e317150095e2e4619caad/gerrychain/partition/assignment.py#L70-L71

PR #384, opened 2022-01-19
Cache flow_from_changes

InnovativeInventor:

### DIFF
--- a/gerrychain/partition/partition.py
+++ b/gerrychain/partition/partition.py
@@ -78,7 +78,7 @@ class Partition:
         self.graph = parent.graph
         self.updaters = parent.updaters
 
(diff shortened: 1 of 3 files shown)


gabeschoenbach (approved):
This all makes sense, my one question is in the new assignment of `self.flows` at line 81:
```python
self.flows = flows_from_changes(parent, self)  # careful
```
`self` here refers to the partition, yes? It's okay to not be the first parameter? I see that when it's called in line 111 of `tally.py` you explicitly pass in a `partition`, so I guess I'm wondering why it necessarily needs to be called `self` in the first place.

InnovativeInventor:
Dicts can't be hashed in Python (so it can't be cached if a `flips` object is passed, which was the original behavior). Partitions, on the other hand, can.

PR #386, opened 2022-01-19
Frozen graph implementation!

InnovativeInventor:

### DIFF
--- a/gerrychain/graph/graph.py
+++ b/gerrychain/graph/graph.py
@@ -1,8 +1,11 @@
+import functools
 import json
+from typing import Any
 import warnings
 
 import geopandas as gp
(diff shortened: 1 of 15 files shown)


InnovativeInventor:
@gabeschoenbach I just cherry-picked a commit from the `retworkx` branch where I fixed a memory leak that appears in very, very long runs.

InnovativeInventor:
@gabeschoenbach I just implemented all your suggestions. I also re-enabled the caching for the `lookup()` function (I disabled it when I was hunting for memory leaks, but forgot to re-enable it. Before re-enabling it, I tested a 100k run to double-check that caching the lookup function does not cause memory leaks.

PR #389, opened 2022-03-18
Add optional max_attempts argument to bipartition tree funcs

InnovativeInventor:
This patch adds a max_attempts argument to the bipartition_tree funcs in `gerrychain.tree`. If `max_attempts` is None, it will default to the previous behavior (i.e. it will hang if it gets stuck). If `max_attempts` is set, it will throw a RuntimeError when `max_attempts` is reached.

This patch is based off code from Robi Huq (via Jeanne Clelland), but is rewritten (to maintain backwards compatibility).

Any mistakes are likely my own.

### DIFF
--- a/gerrychain/tree.py
+++ b/gerrychain/tree.py
@@ -1,6 +1,7 @@
 import networkx as nx
 from networkx.algorithms import tree
 
+from functools import partial
 from .random import random
 from collections import deque, namedtuple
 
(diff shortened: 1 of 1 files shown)


InnovativeInventor:
I also set the default `max_attempts` for the recursive seed partition functions to be 50. This stops them from hanging forever (i.e. getting stuck).

ccthegreat313:
This definitely looks good to me. I'm wondering if 50 max attempts is based on any testing, or if it's a threshold based on logical reasoning? Either way is good, I just don't want 50 to be too limiting nor too time consuming if after, 30 attempts, for example, it becomes a lot less likely for `recursive_seed_part` to return.

InnovativeInventor:
I haven't done any testing. I'm not sure whether Jeanne or Robi Huq did any testing (this default comes from their code). 

It seemed to be a little on the small side, as a default, for me (I personally just bumped it up to 1000 on my local machine) -- bumping it up seems to be very cheap!

ccthegreat313 (approved):

PR #391, opened 2022-04-02
Fix default updaters bug in GeographicPartition

InnovativeInventor:
Several GerryChain users have been reporting a bug in the most recent version of GerryChain where the default updaters in a `GeographicPartition` (e.g. `area`, `boundary_nodes`, etc.) are not being instantiated properly. This PR fixes the bug and adds regression tests to ensure that this doesn't happen again.

It also renames the `use_cut_edges` flag to `use_default_updaters`, which more accurately reflects its intended purpose.

### DIFF
--- a/gerrychain/partition/partition.py
+++ b/gerrychain/partition/partition.py
@@ -31,9 +31,11 @@ class Partition:
         '_cache'
     )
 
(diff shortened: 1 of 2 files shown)


gabeschoenbach (approved):
Looks good!

PR #395, opened 2022-05-20
Fix duplicate test name for geographic partition default updaters

InnovativeInventor:
`mypy` detected a duplicate function definition, which this PR fixes by renaming the duplicate test name.


InnovativeInventor:
Merged due to this PR being a trivial fix.

PR #398, opened 2022-05-23
Improve docs

InnovativeInventor:


InnovativeInventor:
The docs were not building properly/fully because the dependencies were not being properly installed. This fixes it. It also downgrades the PyGeos `ImportError` to a user warning so the docs can still build properly.

InnovativeInventor:
The docs were not building properly/fully because the dependencies were not being properly installed. This fixes it. It also downgrades the PyGeos `ImportError` to a user warning so the docs can still build properly.

PR #400, opened 2022-05-26
Implement diversity stats collector (collect_diversity_stats func)

InnovativeInventor:
This is a draft PR which adds a `collect_diversity_stats()` function that reports the diversity of a given chain. I will come back and add docs/tests.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -92,3 +92,10 @@ Metrics
 
 .. automodule:: gerrychain.metrics
     :members:
+
+
+Diversity stats
+---------------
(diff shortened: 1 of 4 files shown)


InnovativeInventor:
I just added docs and tests. Should be ready for review now.

ccthegreat313 (approved):
This is awesome! I've often wondered about this, given 100,000 steps I feel like it makes more sense to report the unique and accepted plans rather than making it seem like there's definitively 100,000 to report stats on.

PR #401, opened 2022-05-31
Fix uniform spanning tree choice func

InnovativeInventor:
This PR actually fixes two bugs. First of all, since `node_indices` is a set, `random.choice` does not work on it (unless it's first cast to a list). Secondly, sometimes a pure `networkx` graph is passed to `PopulationGraph`, so we cannot assume that the `node_indicies` property exists.


gabeschoenbach:
This looks great! I'll merge it in.

PR #402, opened 2022-06-01
Add initial shims

InnovativeInventor:
This is a PR to add some shims to alternate implementation/sampling methods. These shims can be called from GerryChain Python and behave as a drop-in replacement for normal chain running. The goal is to add shims for:
- [x] frcw
- [ ] GerryChain.NET (maybe)
- [ ] redist (maybe)
- [ ] GerryChain Julia (maybe)

## Example usage
```python
from gerrychain.shims.frcw import frcw_recom, frcw_reversible_recom

# generate seed partition, etc

for partition in frcw_recom(partition, "TOTPOP", ideal_pop, 0.01, steps=10000):
    # normal chain stuff here
```

### DIFF
--- a/gerrychain/shims/__init__.py
+++ b/gerrychain/shims/__init__.py
@@ -1,0 +1,3 @@
+from .frcw import frcw_recom, frcw_reversible_recom
+
+__all__ = ["frcw_recom", "frcw_reversible_recom"]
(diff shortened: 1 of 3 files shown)


pjrule:
Closing this for now—we are planning to provide some form of PyO3 bindings for frcw.rs (indeed, some progress has already been made in this direction), which is significantly less brittle than calling the CLI via `subprocess`.

PR #405, opened 2022-10-29
Port GerryChain adjacency calcs to Shapely 2.0

### LINKED ISSUE #287: Use `maup` for adjacencies?

### LINKED ISSUE #392: Shapely deprecation warning
Graph raises the following deprecation warning when assigning ids to geometries for purposes of calculating adjacency:

```
... /gerrychain/graph/adjacency.py:21: ShapelyDeprecationWarning: Setting custom attributes on geometry objects is deprecated, and will raise an AttributeError in Shapely 2.0
  geometries[i].id = I
```

Another method of tracking geometry identity should be developed before any migration to Shapely 2.0 is undertaken. Perhaps a dict?

InnovativeInventor:
Shapely 2.0 will introduce breaking changes for GerryChain. In particular, our adjacency calculation code will break (and we won't be able to make new dual graphs). Although Shapely 2.0 hasn't stabilized yet, it appears that recent updates in our dependencies (i.e. geopandas) has caused some issues to show up prior to Shapely 2.0 being offically stable.

Relevant issues:
- Fixes #392
- Closes #287

#369 was originally a stop-gap fix for this PR. We should be able to remove the import warning in `gerrychain/__init__.py`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,7 @@ jobs:
             conda install pytest pytest-cov
             conda install codecov
             python -m pip install .[geo]
-            conda install shapely==1.6.4
+            conda install shapely>=2.0.1
 
             # run tests
             echo "backend: Agg" > "matplotlibrc"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,6 @@ jobs:
             conda install pytest pytest-cov
             conda install codecov
             python -m pip install .[geo]
-            conda install shapely>=2.0.1
 
             # run tests
             echo "backend: Agg" > "matplotlibrc"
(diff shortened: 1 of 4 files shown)


InnovativeInventor:
Note: tests are failing due to Shapely 2.0 not being stable yet. Do not merge this PR. We should also add some dependency constraints to ensure that if GerryChain gets updated in a user's install, Shapely will get upgraded as well.

To be clear: this is a breaking change. This PR aims to deprecate Shapely < 2.0.

InnovativeInventor:
We should port `maup` to use GerryChain's adjacency calculations (or vice-versa) so `maup` does not suffer from the same impending deprecation issue.

PR #406, opened 2022-11-17
Use node_repeats in _bipartitions_tree_random_all

### LINKED ISSUE #404: node_repeats Ignored/Unused in bipartition_tree_random
The `node_repeats` parameter of `bipartition_tree_random` is documented as `A parameter for the algorithm: how many different choices of root to use before drawing a new spanning tree.`. However, in fact this parameter has no effect, because it is unused.

The simplest way to observe this is to look at the code and observe that `node_repeats` isn't used:

https://github.com/mggg/GerryChain/blob/3179ac1a4a072377e0d61b4de796cbe5985d5803/gerrychain/tree.py#L224-L257

Instead, a new spanning tree is generated on each attempt (the equivalent of the expected behavior of `node_repeats=1`).

The solution is to simply implement the logic for tracking the number of restarts that is used in the non-randomized `bipartition_tree`:

https://github.com/mggg/GerryChain/blob/3179ac1a4a072377e0d61b4de796cbe5985d5803/gerrychain/tree.py#L206-L219

mkarrmann:
Fixes #404

### DIFF
--- a/gerrychain/tree.py
+++ b/gerrychain/tree.py
@@ -242,17 +242,22 @@ def _bipartition_tree_random_all(
         spanning_tree = spanning_tree_fn(graph)
 
     repeat = True
+    restarts = 0
     attempts = 0
     while max_attempts is None or attempts < max_attempts:
-        spanning_tree = spanning_tree_fn(graph)
+        if restarts == node_repeats:
+            spanning_tree = spanning_tree_fn(graph)
+            restarts = 0
         h = PopulatedGraph(spanning_tree, populations, pop_target, epsilon)
         possible_cuts = balance_edge_fn(h, choice=choice)
 
         repeat = repeat_until_valid
-        attempts += 1
 
         if not (repeat and len(possible_cuts) == 0):
             return possible_cuts
+
+        restarts += 1
+        attempts += 1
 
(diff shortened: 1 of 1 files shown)


mkarrmann:
FYI @pjrule

mkarrmann:
Actually I don't see a way to test this change that wouldn't be tremendously hacky. I'm generally in favor of strong test coverage, but this strikes me a straightforward example of expecting a test would do more harm than good. Do others agree? Are these test checks overridable?

Of course I'm happy to follow any and all policy, I just don't want to push a garbage test just to get around CI. I can also push a test that isn't directly related to my changes if that would satisfy test coverage concerns.

mkarrmann:
Gentle ping on this :) No rush of course.

As I say above, please let me know whether adding a test which covers my diff is a hard requirement or can be overridden. I'm more than happy to abide by any and all policies, I just don't know for certain what they are.

pjrule:
@mkarrmann Yes, you're right that this is an obvious improvement that would be difficult to effectively unit test. I'm merging—thanks for the contribution, and sorry for the long delay!

mkarrmann:
@pjrule Thank you!!

PR #410, opened 2023-05-23
ties are not wins

fgregg:
if two parties tie for first place, the current code will count both of them as winners. this code changes the logic so that a party is a winner if and only if it has more votes than all other opponents.


peterrrock2 (approved):
Looks good! Thanks for catching this!

Issue #414, opened 2023-10-03
recursive_seed_part_inner does not respect provided method


mkarrmann:
`recursive_seed_part` and `recursive_seed_part_inner` accept a parameter `method` which is supposed to be the funciontion used to generate a single district at a give step. However, this `method` is not passed to `recursive_seed_part_inner`'s recursive calls to itself, meaning it is only used to generate the first district, but not any subsequent one.

It is easy to reproduce and observe this bug by calling `recursive_seed_part` with a `method` which has a noticeable side-effect (e.g. logging), and then noticing that that side-effect only occurs once. However, it probably easy to just eye-ball the code to see the problem:

https://github.com/mggg/GerryChain/blob/04ff4f9e8454f5a2216ac6a1c7309dae481ac067/gerrychain/tree.py#L524-L534

https://github.com/mggg/GerryChain/blob/04ff4f9e8454f5a2216ac6a1c7309dae481ac067/gerrychain/tree.py#L591-L605

https://github.com/mggg/GerryChain/blob/04ff4f9e8454f5a2216ac6a1c7309dae481ac067/gerrychain/tree.py#L621-L629

Clearly, `method` is used exactly once, and will not be used in subsequent calls because it is not passed. Instead, the default of `partial(bipartition_tree, max_attempts=10000),` will be used.

PR #415, opened 2023-10-03
#414 Use method in recursive calls to recursive_seed_part_inner

mkarrmann:
See https://github.com/mggg/GerryChain/issues/414 for description of issue.

The solution is self-explanatory.

### DIFF
--- a/tests/test_tree.py
+++ b/tests/test_tree.py
@@ -120,7 +120,11 @@ def test_recursive_seed_part_uses_method(twelve_by_twelve_with_pop):
         ceil=None,
         method=dummy_method,
     )
-    assert calls == n_districts - 1
+    # Called at least once for each district besides the last one
+    # (note that current implementation of recursive_seed_part calls method
+    # EXACTLY once for each district besides the last one, but that is an
+    # implementation detail)
+    assert calls >= n_districts - 1
 
 def test_random_spanning_tree_returns_tree_with_pop_attribute(graph_with_pop):
     tree = random_spanning_tree(graph_with_pop)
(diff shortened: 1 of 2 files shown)


mkarrmann:
Something weird is going on with pip install in CI, will have to investigate later...

mkarrmann:
Not sure what's going on with `codecov`, I'm clearly not decreasing test coverage here.

mkarrmann:
@InnovativeInventor @pjrule FYI, a very simple bug fix. A unit test to test the change is included.

tests/test_tree.py
```diff
@@ -87,6 +87,41 @@ def test_recursive_seed_part_returns_within_epsilon_of_target_pop(twelve_by_twel
...
+        twelve_by_twelve_with_pop,
+        range(n_districts),
+        ideal_pop,
+        "pop",
+        epsilon,
+        n=5,
+        ceil=None,
+        method=dummy_method,
+    )
+    assert calls == n_districts - 1
```
InnovativeInventor, line 123: This assertion is a little brittle and might break if the internal implementation of `recursive_seed_part` changes in the future. I think it's alright to have here, though.

InnovativeInventor (approved):
These changes make sense to me. 

I'm not sure why codecov is being weird about this and flagging those lines in particular. Maybe something changed in codecov?

tests/test_tree.py
mkarrmann: @InnovativeInventor Ty, great call out.

I updated it to be a `>=` check, since I think there's a stronger case to be made that `method` is called *at least* once is part of `recursive_seed_part`'s contract, let me know your thoughts.

Tbh I'd also be fine scrapping the unit test if you prefer, considering a regression here doesn't seem terribly likely.

InnovativeInventor: Great, LGTM.

InnovativeInventor (approved):

PR #436, opened 2026-01-19
Refactoring of tree.py and additional cleanup of alpha code

chief-dweeb:
To be honest, I forget everything that is in this PR, but what I have just wrapped up is a major refactoring of tree.py into new files nested under a new directory "tree".  There is a block comment at the top of tree/bipartition_tree.py that describes the refactoring.


gerrychain/proposals/tree_proposals.py
```diff
@@ -36,14 +37,109 @@ class ValueWarning(UserWarning):
     pass
 
 
+# frm: Note to Peter:  param name, "method", has been changed everywhere to "bipartition_tree_fn"
+#
+# I have added a note to the rx_release_notes.md file that we should warn users about this
+# change.
```
peterrrock2: 👍

gerrychain/proposals/tree_proposals.py
```diff
@@ -368,7 +481,15 @@ def __init__(
...
-        method: Callable = bipartition_tree_random,
+        bipartition_tree_fn: Callable = bipartition_tree,
+        # frm: TODO: Refactoring: Delete this comment after Peter OK's PR
+        #
+        # I changed this from biparition_tree_random to bipartition_tree because every other
+        # invocation of recom() that I saw passed it bipartition_tree(), so I assumed that this
+        # use of bipartition_tree_random was left over from pre-history.
+        #
+        # Note that I have not been able to find any use of the ReCom class, so I think we
+        # should just delete it entirely...
```
peterrrock2: I am not so sure that we want to get rid of this. I think that this was built by someone trying to solve the problem where, when a user wants to run a chain with ReCom, they have to do the following bit of syntax twister:

```python
from functools import partial

proposal = partial(
    recom,
    pop_col="TOTPOP",
    pop_target=ideal_population,
    epsilon=0.01,
    node_repeats=2
)
```

This partial application is familiar to anyone that does functional programming, and is effectively just

```python
def proposal(state: Partition) -> Partition:
    return recom(
        state,
        pop_col="TOTPOP",
        pop_target=ideal_population,
        epsilon=0.01,
        node_repeats=2,
    )
```

in disguise. But our users are not programmers, and I get a lot of questions about the "partial" function that appears in the documentation. The `ReCom` class does this partial application under the hood using its `__call__` attribute, and eliminates the need for the import of `partial` from `functools`, so the user only has to do 

```python
Recom(
    pop_col="TOTPOP",
    ideal_pop=ideal_population,
    epsilon=0.01,
)
```

rather than the `partial` rigmarole.

I discovered this class existed in the codebase after doing a big refresh on the main documentation a while ago. I have been waiting to update things until a major release because the values that need to be passed to the `__init__` of `Recom` depend on the underlying bipartition function, and the class would probably be better transformed into something closer to a "namespace" to improve discoverability


```python
class ReCom:
    def __init__(self, *args, **kwargs):
        raise TypeError("ReCom is not instantiable; use ReCom.mst(...), etc.")

    @classmethod
    def mst(...):
        # minimum spanning tree version

    @classmethod
    def B(...):
        # This is the "district pairs minimum spanning tree".  An absolutely terrible name 
        # for a function, to be sure, but it will make replication of what is in the 
        # Reversible ReCom paper (https://data-democracy.org/rrc) we published a while
        # back easier for other people. This function would just be a wrapper around mst
        # defined above.
```

and so on. I am happy to put all of this functionality in later if you don't want to mess with it.

gerrychain/chain.py
```diff
@@ -76,6 +76,13 @@ def __init__(
 
         :raises ValueError: If the initial_state is not valid according to the constraints.
         """
+
+        # frm: TODO: Refactoring: Is it worth it to investigate whether the proposal function makes sense?
+        #
+        # I just feel a little odd about accepting a Callable with no checking that it does what it is supposed
+        # to do.  I know that this sounds like a broken record, but when a function parameter does the wrong
+        # thing it can be hell to debug...
+
```
peterrrock2, line 85: It's probably worth checking this in the init. Right now, we should just update the signature to be more
explicit about the typing:

```python
proposal: Callable[[Partition], Partition]
```

and add a smoke test

```python
def _smoke_test_proposal(proposal: Callable[[Partition], Partition], initial_state: Partition) -> None:
    try:
        next_state = proposal(initial_state)
    except Exception as e:
        raise TypeError(
            f"proposal(initial_state) raised {type(e).__name__}: {e}"
        ) from e

    if not isinstance(next_state, Partition):
        raise TypeError(
            f"proposal function must return a Partition ; got {type(next_state)!r}"
        )
```

In the long run, it's probably best to add a type alias so that we don't have to keep typing `Callable[[Partition], Partition]`. Something like

```python
from typing import TypeAlias, Callable

ProposalFn: TypeAlias = Callable[[Partition], Partition]
```

or we can use a protocol (a.k.a. interface) if we want to add a doc string to give users context:

```python
from typing import Protocol, runtime_checkable

# runtime_checkable just allows you to use isinstance for quick checks to make sure 
# that a supposed ProposalFn is, in fact, callable (has the __call__ attribute)
@runtime_checkable
class ProposalFn(Protocol):
    """
    A function that takes in a Partition object (state) and returnd the next Partition (state) 
    in the Markov chain.
    """
    def __call__(self, state: Partition) -> Partition: ...
```

gerrychain/tree/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+of those files for more information.
+"""
+
+# frm: TODO: Documentation:  Update documentation for tree/__init__.py
+#
+# It is just a copy of the __init__.py from graph, so it needs to
+# be changed to apply to tree...
+
+from .bipartition_tree import *
+from .spanning_tree import *
```
peterrrock2: Please change these to explicit imports and put things intended to be part of the public API into an `__all__` list before you are done.

gerrychain/tree/bipartition_tree.py
```diff
@@ -190,264 +184,18 @@
...
-    # Pick a starting point at random
-    root_id = choice(list(graph.node_indices))
-
-    # Initiallize the tree to contain the root_node (with no parent)
-    tree_nodes = set([root_id])
-    parent_node_id = {root_id: None}
+# frm: Note to Peter - moved random_spanning_tree() to tree/spanning_tree.py
 
-    for node_id in graph.node_indices:
+# frm: Note to Peter - moved uniform_spanning_tree() to tree/spanning_tree.py
```
peterrrock2: 👍

gerrychain/tree/bipartition_tree.py
```diff
@@ -600,48 +348,53 @@ def __repr__(self) -> str:
...
+#
+# Peter said (January 2026): I am fine with changing the name of that parameter.
+# Python has good error messages for when things have the wrong name:
+# and if users pass this as a positional argument, then it won't matter that the
+# parameter name changed.
 
 
 def find_balanced_edge_cuts_contraction(
-    h: PopulatedGraph, one_sided_cut: bool = False, choice: Callable = random.choice
+    h: _PopulatedGraph, one_sided_cut: bool = False, root_choice_fn: Callable = random.choice
```
peterrrock2:
```suggestion
    h: _PopulatedGraph, one_sided_cut: bool = False, rootnode_choice_fn: Callable = random.choice
```

Sorry, small nit-pick here. Can we use this name instead?

gerrychain/tree/bipartition_tree.py
```diff
@@ -190,264 +184,18 @@
...
-    :rtype: Graph
-    """
-
-    # Pick a starting point at random
-    root_id = choice(list(graph.node_indices))
-
-    # Initiallize the tree to contain the root_node (with no parent)
-    tree_nodes = set([root_id])
-    parent_node_id = {root_id: None}
+# frm: Note to Peter - moved random_spanning_tree() to tree/spanning_tree.py
```
peterrrock2: :+1:

gerrychain/tree/bipartition_tree.py
```diff
@@ -1280,10 +1091,13 @@ def _region_preferred_max_weight_choice(
 #      GerryChain does and how it does it.  Everything else is just flow control and
 #      book-keeping
 #
-# Get Peter's feedback on whether this all makes sense.
+#        => This is essentially DONE.  The logic for bipartition_tree for all cases is
+#           now encapculated in two routines, _internal_bipartition_tree() and
+#           _get_possible_edge_cuts_and_populated_graph().
```
peterrrock2: :+1:

gerrychain/tree/bipartition_tree.py
```diff
@@ -1369,139 +1183,99 @@ def old_bipartition_tree(
...
+        spanning_tree_fn=spanning_tree_fn,
+        find_balanced_edge_cuts_fn=find_balanced_edge_cuts_fn,
+        root_choice_fn=root_choice_fn,
+        max_attempts=max_attempts,
+        warn_attempts=warn_attempts,
+        repeat_until_valid=repeat_until_valid,
+        allow_pair_reselection=allow_pair_reselection,
+    )
+    possible_cuts = cuts_and_populated_graph[0]
+    populated_graph = cuts_and_populated_graph[1]
```
peterrrock2:
```suggestion
    possible_cuts, populated_graph = _get_possible_edge_cuts_and_populated_graph(
        graph_to_split=subgraph_to_split,
        pop_col=pop_col,
        pop_target=pop_target,
        epsilon=epsilon,
        node_repeats=node_repeats,
        spanning_tree=spanning_tree,
        spanning_tree_fn=spanning_tree_fn,
        find_balanced_edge_cuts_fn=find_balanced_edge_cuts_fn,
        root_choice_fn=root_choice_fn,
        max_attempts=max_attempts,
        warn_attempts=warn_attempts,
        repeat_until_valid=repeat_until_valid,
        allow_pair_reselection=allow_pair_reselection,
    )
```

gerrychain/tree/bipartition_tree.py
```diff
@@ -1369,139 +1183,99 @@ def old_bipartition_tree(
...
+            raise Exception("This should never happen...")
 
 
-##################################
-# Start of new code:
+# frm: TODO: Refactoring: Delete old_bipartion_tree() after Peter has done PR
+#
+# Note to Peter:  I am including this code so that you can have it at hand to
+# see what used to happen before I unified the logic for all of the
+# different bipartition_tree functions.
```
peterrrock2: I appreciate that. Thank you!

gerrychain/tree/bipartition_tree.py
```diff
@@ -1630,42 +1390,53 @@ def bipartition_tree(
...
+            and "populated_graph" in signature(cut_choice_fn).parameters
         )
 
         chosen_cut = None
         if is_region_cut:
-            chosen_cut = cut_choice(populated_graph, region_surcharge, possible_cuts)
+            chosen_cut = cut_choice_fn(populated_graph, region_surcharge, possible_cuts)
         else:
-            chosen_cut = cut_choice(possible_cuts)
+            chosen_cut = cut_choice_fn(possible_cuts)
```
peterrrock2: Yeah, I think that we need make a more forcing contract for what the signature of the cut_choice_fn looks like. Probably something like

```python
def example_cut_choice_fn(
    possible_cuts: Sequence[Cut],
    **kwargs
)
```

which we can spell out in a protocol (a.k.a. interface)

```python
from typing import Protocol, Sequence, Any

class CutEdgeChoiceFn(Protocol):
    def __call__(self, possible_cuts: Sequence["Cut"], **kwargs: Any) -> "Cut": ...
```

This way, if you pass kwargs that are not used, then they just do nothing, but we can have a bunch of different cut choice functions that will work in this section

```python
def random_cut(possible_cuts: Sequence[Cut], **kwargs: Any) -> "Cut":
    return random.choice(possible_cuts)

def region_preferred_max_weight_choice(
    possible_cuts: Sequence[Cut], 
    * , 
    populated_graph: PopulatedGraph, 
    region_surcharge: Dict, 
    **kwargs: Any
) -> Cut:
     # existing code goes here
```

which both satisfy the protocol. This doesn't really do a lot at runtime to protect users, but it does help python typecheckers make sure you haven't done anything illegal.

I'm not sure if the above is better, or if it would be less confusing to end users to just force them to call a function that looks exactly like

```python
def example_function_with_binding_contract(
    possible_cuts: Sequence[Cut], 
    * , 
    populated_graph: PopulatedGraph, 
    region_surcharge: Dict, 
) -> Cut:
```

and to just tell them that sometimes those extra arguments are not used. I am open to input

gerrychain/tree/bipartition_tree.py
```diff
@@ -2090,867 +1950,151 @@ def old_bipartition_tree_random(
...
 #
-#   2) It will declutter tree.py which is already complex and hard to understand
+# So the implementation of bipartition_tree_random() is now just a call on bipartition_tree().
 #
-# This may cause legacy users some pain if they use these routines in code that does
-# not already import partition, but the error message should be clear enough to let them
-# figure it out easily.
+# Once Peter approves, I will delete this.
+#
+# This should be added to the release notes just in case someone in the wide world uses it.
```
peterrrock2: I'm fine with getting rid of it. It looks like the only other difference was in what the default "cut_choice_fn" was

gerrychain/partition/initial_partition_generators.py
```diff
@@ -0,0 +1,718 @@
...
+
+
+    Also - address this comment which I now do not grok:
+
+        OK, but why is the logic above for num_chunks the correct number?  Is there
+        a mathematical reason for it?  I assume so, but that explanation is missing...
+
+        I presume that the reason is that something in the code that finds a
+        district scales exponentially, so it makes sense to divide and conquer.
+        Even so, why this particular strategy for divide and conquer?
```
peterrrock2, line 490: The reason that we needed `recursive_seed_part` was because `recursive_tree_part` can paint you into a metaphorical corner sometimes. There are 2 main issues with the `recursive_tree_part` that appear when you are trying to subdivide into a large number of districts that the seed part method solves

1. `recursive_tree_part` only carves off one district at a time,  but since each of the districts are approximately population balanced, you need to dynamically adjust the target range as you carve things off. If you cut off something slightly larger than ideal, then adjust the target for the next district down, and visa versa. This sort of bookeeping is a bit tricky, and, if the allowable population band is too large, can end up with you getting stuck.

2. Since `recursive_tree_part` only carves off one district at a time, if there are a lot of districts to cut off, you can cut off districts in such a way that the remaining graph is impossible to population balance. This is mostly caused by cutting off districts in a "snake-y" way so as to leave very small bits of unassigned graphs (with possibly high nodes population) between assigned areas. In the abstract, you can imagine that you produce something very close to a path with the following populations:

```
4 - 10 - 12 - 8
``` 

There is no way to subdivide this graph into 2 pieces of even size. In fact, you can't even get within 10% population tolerance (17 +/- 1.7). 

The `recursive_seed_part` method keeps this from happening so much by biting off large, approximately balanced things, and then recursing inward. Under this schema, it's a lot harder to make a snake-y patchwork of districts.

peterrrock2 (commented):
Great work on this! Let me know when you have had a chance to look at all of the comments, and then we'll get this merged in.

gerrychain/tree/bipartition_tree.py
chief-dweeb: I removed this comment from the code

gerrychain/tree/bipartition_tree.py
chief-dweeb: I deleted bipartition_tree_random() and all references to it.

gerrychain/proposals/tree_proposals.py
chief-dweeb: Makes sense.  I added a comment in the code (inserting what you said above) along with some comments of my own.  I added a snippet of code for what the resulting class would look like - if you agree, then I will add that code and then let you augment it in the future with other recom methods you want to add to the namespace...

gerrychain/tree/__init__.py
chief-dweeb: DONE

gerrychain/tree/bipartition_tree.py
chief-dweeb: DONE

gerrychain/tree/bipartition_tree.py
chief-dweeb: I deleted the old_bipartition_tree() and old_bipartition_tree_with_num_cuts() now that you have done the PR

gerrychain/partition/initial_partition_generators.py
chief-dweeb: Makes sense - I added your explanation to the top-level docstrings in the file.

peterrrock2:
I will be adding notes to this comment as we go through user sessions. For now, I will just bullet API inconsistencies from the migration, but will later include more user feedback
- `uniform_spanning_tree` change parameter `choice` -> `root_choice_fn`
- `reversible_recom` change paramter `choice` -> `root_chice_fn`

PR #437, opened 2026-02-24
Documentation update for rustworkx migration

chief-dweeb:
This is almost all documentation changes.  I wanted to get them in before you (Peter) did the docstring overhaul.

There is one question in the code that needs to be answered for the documentation to be accurate - the question is about how region_surcharge works in random_spanning_tree().  Since region_surcharge is so useful, I think it makes sense to be crystal clear about how it works (and when it doesn't work).

### DIFF
--- a/gerrychain/tree/spanning_tree.py
+++ b/gerrychain/tree/spanning_tree.py
@@ -11,161 +11,69 @@ from ..graph import Graph
 # frm: TODO: Documentation: Update the high level description for spanning_tree.py below
 
 """
-This module provides tools and algorithms for manipulating and analyzing graphs,
-particularly focused on partitioning graphs based on population data. It leverages the
-GerryChain Graph object to handle graph structures and implements various algorithms for graph
-partitioning and tree traversal.
+This module provides two implementation of spanning tree functions.
+
+A spanning tree is a tree formed from a connected undirected graph that contains
+all of the nodes in the graph, but only some of the edges, so that the resulting
+graph is a tree (with no cycles). You can learn more here:
+
+    https://en.wikipedia.org/wiki/Spanning_tree
+
+Given an spanning tree for a graph, one can easily compute the population of every
+subtree by doing a bottom up tree traversal.  Given this, one can carve off subtrees
+that have the desired population, one by one, and thereby construct a set of
+districts (note that this can fail if the population of a subtree is below epsilon
+of the population target and the parent subtree is above epsilon).
+
+Two implementations are provided:
+
+    A uniform spanning tree:
+
+        A spanning tree chosen randomly from among all the spanning trees with equal
+        probability is called a uniform spanning tree. Wilson's algorithm can be
+        used to generate uniform spanning trees in polynomial time by a process
+        of taking a random walk on the given graph and erasing the cycles created
+        by this walk.
+
+    A minimum spanning tree (MST):
+
+        Unlike a uniform spanning tree, an MST selects a spanning tree from
+        a subset of all spanning trees, based on edge weights.  It selects
+        a spanning tree that minimizes the total of all of the edge weights
+        in the graph.
+
+        In GerryChain, the random_spanning_tree() function implements an
+        MST using Kruksal's Algorithm along with random weights.  That is,
+        the code associates a random weight (between 0 and 1) to each
+        edge in the graph and then computes the MST for those weights.
+
+        The random_spanning_tree() function additionally supports
+        a "region surcharge" that increases the weights for specific
+        edges.  This allows one to construct a spanning tree that
+        tends to keep surcharged edges together in the same district.
+        The details are documented elsewhere - the point here is that
+        the random_spanning_tree() function is the one you want to use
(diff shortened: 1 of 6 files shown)


gerrychain/tree/spanning_tree.py
```diff
@@ -193,30 +117,58 @@
...
+# spanning tree, because the find_balanced_edge_cuts functions all go
+# bottom up.
+#
+# In short, I find that I do not really grok why region surcharge works,
+# how effective it really is, and why there is not a better solution to
+# keeping regions together...
+#
+# Another issue is how users should think about the weights used for
+# a region surcharge.  This is especially interesting when there are
+# more than one "region" being surcharged - for instance "muni" and "water".
```
peterrrock2, line 155: The `region_surcharge` parameter is a clever trick to modify the way that Kruskal's samples from the space of all possible spanning trees.

For the sake of simplicity, imagine that we have a set regions that tile our map (e.g. counties). Then the `region_surcharge` parameter will look at the base graph and assign a surcharge weight, say 0.3 for sake of example, to edges between distinct regions of the same type. So if there is an edge between a node in County A and County B, then, when we go to roll the random weight for that edge for Kruskal's (a value between 0 and 1), we will add the surcharge weight to that edge ex-post-facto. This has the effect of making edges between regions less likely to be selected early on by Kruskal's algorithm. 

Pushing this to the extreme, it is not too hard to see if you assign a surcharge weight of 1.0, then at some point in Kruskal's algorithm you are forced into a spanning forest where each tree in the forest is it's own region. In fact, if there are $n$ nodes in the graph and $k$ regions, then this forest appears on step $n-k$ of  Kruskal's algorithm. With the 1.0 surcharge, that means that the remaining $k-1$ edges that need to be selected for the spanning tree can only be between your chosen regions. 

Then comes the population balancing step. Since you only have one edge in between each region of interest when the surcharge is 1.0, when you go to bisect the tree, you are then FORCED to leave at least  $k-1$ of the regions of interest whole. Note that this is not at all guaranteed when there is no surcharge on edges between regions since trees in the spanning forest are then free weave in and out of regions at their leisure. 

In this way, you can think of  `region_surcharge` as a parameter that modifies the graph partition that appears at step $n-k$ of Kruskal's algorithm where a higher value increases the probability of seeing a spanning forest on the regions of interest. As the value decreases from 1 to 0, you increase the amount of "weaving" between regions that is allowed at step $n-k$ of Kruskals, and, therefore decrease the number of regions that you expect to keep whole.

When there are multiple types of region with a surcharge of 1.0, the idea is the same, but the exact math of it gets a little messy since you need to consider intersections. TL;DR is that there will be some step in Kruskal's algorithm where you have $\ell$ pieces of partition where each piece is a member of the (topological) common refinement of the region types, and then you build a spanning tree on that. In this case, the number of pieces that you are guaranteed to keep whole depends entirely on the way that the region types are laid out, and there are ways of arranging regions that would force you into a situation with no guarantees (i.e. a grid where the types of region are "rows" and "columns" since the common refinement has $n\cdot m$ pieces -- the individual nodes -- and so you are back to basic Kruskal's).

gerrychain/tree/spanning_tree.py
```diff
@@ -244,18 +196,17 @@ def random_spanning_tree(graph: Graph, region_surcharge: Optional[Dict] = None)
...
+    #
+    # We should make clear what the issues are for different values of
+    # region_surcharge.  The random values are between 0-1, so any region
+    # surcharge value greater than 1 will dominate.  What happens if the
+    # user provides several region surcharge values for different
+    # node attributes?  I do not know, and I presume users won't know.
+    #
+    # Note that the documentation currently existing sometimes says that
+    # region surcharge values should be between 0-1 and then it has
+    # example code where the values are 1 or greater than 1...
```
peterrrock2, line 209: Noted. Values above can be useful since then you can place an importance ordering on the common refinement, but for most people, once the values get above 1, there is not a whole lot of utility.

gerrychain/tree/spanning_tree.py
```diff
@@ -327,6 +278,31 @@ def random_spanning_tree(graph: Graph, region_surcharge: Optional[Dict] = None)
...
+            #
+            # But the code below surcharges both when the nodes are in different regions
+            # AND when either of the nodes is not in a region at all.  From what the
+            # article says, we should not surcharge when BOTH nodes are not in a region
+            # at all.
+            #
+            # ...confused...
+            #
+            # Ask Peter...
+            #
```
peterrrock2, line 304: This is something that we have gone back and forth on in the lab. There are meaningful differences in whether you should treat members of the "void" (places without a region assignment) as all being members of some "void" region or as individual regions unto themselves (the implementation given here). Currently, this is here because it has worked in the past for the outcomes that we desired, but we are actively working on figuring out the best practice here.

What will probably happen is we will introduce another parameter that will allow you to deal with "void" nodes in a variety of ways.

peterrrock2 (approved):

PR #443, opened 2026-05-12
Frm rustworkx post alpha - clean up documentation mostly

chief-dweeb:
Peter - please look for all comments that have your name in them "Peter" that show up in this PR.

I have added a ProposalFn Protocol class to document what a proposal function should look like - takes a partition and returns a partition as a way to clearly indicate what MarkovChain expects.  The question is whether you like that approach and if so, where it should be used - details in the comments in the PR, but the short story is that I have changed MarkovChain to use it, and I am suggesting that we encourage the use of new generate_xxx_protocol() -> ProposalFn functions as the way users obtain a proposal function.  This would just be syntactic sugar for existing flip proposal functions that already only take one parameter (Partition), but it would be the way we handled hidiing making proposals using partial functions.

Lots of comments removed - if I thought that they were long-term interesting, I copied them into a file I am keeping of things we might do in the future.  The reason to defer them is to just get this release out - less is more....

-Fred

### DIFF
--- a/gerrychain/graph/graph.py
+++ b/gerrychain/graph/graph.py
@@ -362,26 +362,6 @@ class Graph:
 
         return nx_graph
 
-    # frm: TODO: Refactoring: Peter: What you think about "cosmetic" type definitions?
-    #
-    # There are several places where the code has type-hints of "Any" or "int".  I would like for
-    # purely stylistic reasons to substitute more informative terms.  For instance, node_ids
-    # can be integers or tuples or strings, so they are currently "typed" as "Any", but we
-    # could define a "node_id_type" that was just syntactic sugar for "Any" and make the
-    # function signatures make more sense.
-    #
-    # These definitions would also provide a convenient and logical place to talk about
-    # the specifics of each "type" - NX vs. RX, districts as part of partitions, etc.
-    #
-    # Specific cases:
-    #
-    #     node_id_type for "Any"
-    #     edge_id_type for "Any"
-    #     edge_type for "tuple[node_id_type, node_id_type]".
-    #     district_id_type for "int"
-    #     flip_dict_type for dict[node_id_type, district_id_type])
-    #
-
     def original_nx_node_id_for_internal_node_id(self, internal_node_id: Any) -> Any:
         """Translate a node_id to its "original" node_id.
 
@@ -1029,10 +1009,6 @@ class Graph:
         """
         # Return a set of edge tuples
 
-        # frm: TODO: Code: ???: Should edges return a list instead of a set?
-        #
-        # Peter said he thought users would expect a list - but why?
-
         self.verify_graph_is_valid()
 
         if self.is_rx_graph():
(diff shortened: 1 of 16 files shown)


gerrychain/graph/graph.py
```diff
@@ -1233,29 +1202,13 @@ def __getattr__(self, __name: str) -> Any:
...
-        # Peter's comments from PR:
-        #
-        # Oh interesting; good catch! The flag approach seems like a good solution to me.
-        # It's very, very rare to use the default constructor, so I don't imagine that
-        # people will really run into this.
 
-        # frm: TODO: Code: Fix this hack (in __getattr__) - see comment above...
+        # Avoid an infinite loop that can happen if a user happens to use the default
+        # constructor for a Graph.  In that case the Graph object will have no
+        # attributes, and that is bad, because the code for is_rx_traph() below
```
peterrrock2:
```suggestion
        # attributes, and that is bad, because the code for is_rx_graph() below
```
nit: typo

gerrychain/proposals/proposals.py
```diff
@@ -1,10 +1,76 @@
...
+#
+#     my_proposal = generate_any_node_flip_proposal()
+#
+#     chain = MarkovChain(
+#       my_proposal,
+#       ...
+#       )
+#
+# So, the question for Peter is whether this sounds like a good idea or just
+# a bunch of syntactic sugar "hot air".
```
peterrrock2: I think that adding the proposal function generators is a very good idea. I would prefer to call them `build_xxx_proposal` as opposed to `generate_xxx_proposal` simply because this would then be a part of the public API, and users generally prefer shorter names for things.

gerrychain/proposals/proposals.py
```diff
@@ -1,10 +1,76 @@
...
+any additional information to be used by a ProposalFn needs to
+be bound using a "partial" function definition.  For instance,
+the recom() function needs to know which bipartition function
+to use, and so the appropriate bipartition function is bound
+using a "partial" function in order to create a ProposalFn
+that only takes the Partition object as its single parameter.
+
+Convenience routines are often provided to do this "partial"
+function binding.
+"""
```
peterrrock2, line 21: I would probably modify the wording to make it 
```suggestion
"""
This module defines proposal functions for use with MarkovChain.

A ProposalFn is a function that takes a Partition as its only argument and
returns a new Partition, and is used by the MarkovChain to generate
the next Partition at each step of the chain.

Because a ProposalFn must accept only a single Partition argument, any
additional information to be used by a ProposalFn needs to be bound 
ahead of time. This is often done by partially applying the proposal 
function with functools.partial, or by using a helper function that returns 
a closure capturing the additional parameters.

For instance, the recom() function needs to know which bipartition function 
to use, and so the appropriate bipartition function is bound in advance through
one of the aformentioned methods and the returned a ProposalFn will have the
expected single-argument signature.

Since using the paritial function or creating an appropriate closure can be 
unintuitive, convenience functions are often provided to perform this binding.
"""
```

gerrychain/proposals/proposals.py
peterrrock2: Thinking about this further, it is probably worth promoting these as the main interaction surface for users.

chief-dweeb: I made a bunch of changes to support ProposalFn as the standard way to get a proposal function, and then pushed the changes, so you should be able to see them now...

peterrrock2 (approved):
This is looking good!

PR #444, opened 2026-06-11
Optimization pass before 1.0.0 release

peterrrock2:
## Summary

Post-migration performance and robustness pass over the rustworkx-backed `Graph`. Removes the hot-path overhead that profiling flagged after the rustworkx migration landed: attribute-delegation cost in `FrozenGraph`, per-call graph validation, subgraph-wrapper construction in the contiguity constraints, and a Dijkstra-where-BFS-suffices in `single_flip_contiguous`. Also addresses several of the hanging migration review concerns mentioned by @chief-dweeb previously.

Improvements here seem to show a ~20-30% speedup over the previous merge (f1a2970fd5df23470a560eba440db2a9e8c624ba). Speed improvements were measured using the new benchmark suite included in this PR.

```bash
uv run python benchmarks/benchmark_recom.py \
  compare pypi:0.3.2 git:f1a2970fd5df23470a560eba440db2a9e8c624ba local \
  --json benchmarks/graphs/<GRAPH> \
  --parts <PARTS> \
  --steps <STEPS>
```

### Results on some graphs

**PA.json: 1000 steps, 17 parts, 5 seeds, 1 trial per seed, ε=0.01**

| GerryChain Version  | Min Observed Time per Step | Max Observed Time per Step | Median Time per Step | Speedup vs. 0.3.2 |
| ------------------- | -------------------------- | -------------------------- | -------------------- | ----------------- |
| 0.3.2 (PyPI)        | 106.535 ms                 | 121.360 ms                 | 119.742 ms           | Baseline          |
| 1.0.0a2 (f1a2970fd) | 58.419 ms                  | 67.605 ms                  | 60.829 ms            | ~1.97×            |
| 1.0.0a2 (local)     | 42.274 ms                  | 48.562 ms                  | 43.837 ms            | ~2.73×            |

**hard_graph.json: 250 steps, 6 parts, 5 seeds, 1 trial per seed, ε=0.01**

| GerryChain Version  | Min Observed Time per Step | Max Observed Time per Step | Median Time per Step | Speedup vs. 0.3.2 |
| ------------------- | -------------------------- | -------------------------- | -------------------- | ----------------- |
| 0.3.2 (PyPI)        | 863.995 ms                 | 1048.825 ms                | 974.904 ms           | Baseline          |
| 1.0.0a2 (f1a2970fd) | 582.955 ms                 | 869.971 ms                 | 661.009 ms           | ~1.47×            |
| 1.0.0a2 (local)     | 491.957 ms                 | 722.360 ms                 | 549.640 ms           | ~1.77×            |

**grid_8x8.json: 10000 steps, 8 parts, 5 seeds, 1 trial per seed, ε=0.01**

| GerryChain Version  | Min Observed Time per Step | Max Observed Time per Step | Median Time per Step | Speedup vs. Baseline |
| ------------------- | -------------------------- | -------------------------- | -------------------- | -------------------- |
| 0.3.2 (PyPI)        | 0.524 ms                   | 0.529 ms                   | 0.526 ms             | Baseline             |
| 1.0.0a2 (f1a2970fd) | 0.369 ms                   | 0.378 ms                   | 0.377 ms             | ~1.39×               |
| 1.0.0a2 (local)     | 0.281 ms                   | 0.288 ms                   | 0.286 ms             | ~1.84×               |

**grid_25x25.json: 10000 steps, 5 parts, 5 seeds, 1 trial per seed, ε=0.01**

| GerryChain Version  | Min Observed Time per Step | Max Observed Time per Step | Median Time per Step | Speedup vs. Baseline |
| ------------------- | -------------------------- | -------------------------- | -------------------- | -------------------- |
| 0.3.2 (PyPI)        | 12.340 ms                  | 12.563 ms                  | 12.384 ms            | Baseline             |
| 1.0.0a2 (f1a2970fd) | 8.422 ms                   | 8.757 ms                   | 8.632 ms             | ~1.43×               |
| 1.0.0a2 (local)     | 5.856 ms                   | 6.081 ms                   | 6.009 ms             | ~2.06×               |

**grid_50x50.json: 1000 steps, 25 parts, 5 seeds, 1 trial per seed, ε=0.01**

| GerryChain Version  | Min Observed Time per Step | Max Observed Time per Step | Median Time per Step | Speedup vs. Baseline |
| ------------------- | -------------------------- | -------------------------- | -------------------- | -------------------- |
| 0.3.2 (PyPI)        | 9.301 ms                   | 10.306 ms                  | 9.626 ms             | Baseline             |
| 1.0.0a2 (f1a2970fd) | 6.287 ms                   | 7.206 ms                   | 6.771 ms             | ~1.42×               |
| 1.0.0a2 (local)     | 4.387 ms                   | 5.074 ms                   | 4.737 ms             | ~2.03×               |

**Actual Outputs:**

```console
==== summary: graph=PA.json, 5 seed(s) x 1 repeat(s), 1000 steps, parts=17, epsilon=0.01, jobs=5 ====
pypi:0.3.2                                                median   119.742 ms/step  [min   106.535, max   121.360]  gerrychain 0.3.2      (baseline)
git:f1a2970fd5df23470a560eba440db2a9e8c624ba (f1a2970fd)  median    60.829 ms/step  [min    58.419, max    67.605]  gerrychain 1.0.0a2    (1.97x baseline speed)
local (working tree)                                      median    43.837 ms/step  [min    42.274, max    48.562]  gerrychain 1.0.0a2    (2.73x baseline speed)

==== summary: graph=hard_graph.json, 5 seed(s) x 1 repeat(s), 250 steps, parts=6, epsilon=0.01, jobs=5 ====
pypi:0.3.2                                                median   974.904 ms/step  [min   863.995, max  1048.825]  gerrychain 0.3.2      (baseline)
git:f1a2970fd5df23470a560eba440db2a9e8c624ba (f1a2970fd)  median   661.009 ms/step  [min   582.955, max   869.971]  gerrychain 1.0.0a2    (1.47x baseline speed)
local (working tree)                                      median   549.640 ms/step  [min   491.957, max   722.360]  gerrychain 1.0.0a2    (1.77x baseline speed)

==== summary: graph=grid_8x8.json, 5 seed(s) x 1 repeat(s), 10000 steps, parts=8, epsilon=0.01, jobs=5 ====
pypi:0.3.2                                                median     0.526 ms/step  [min     0.524, max     0.529]  gerrychain 0.3.2      (baseline)
git:f1a2970fd5df23470a560eba440db2a9e8c624ba (f1a2970fd)  median     0.377 ms/step  [min     0.369, max     0.378]  gerrychain 1.0.0a2    (1.39x baseline speed)
local (working tree)                                      median     0.286 ms/step  [min     0.281, max     0.288]  gerrychain 1.0.0a2    (1.84x baseline speed)

==== summary: graph=grid_25x25.json, 5 seed(s) x 1 repeat(s), 10000 steps, parts=5, epsilon=0.01, jobs=5 ====
pypi:0.3.2                                                median    12.384 ms/step  [min    12.340, max    12.563]  gerrychain 0.3.2      (baseline)
git:f1a2970fd5df23470a560eba440db2a9e8c624ba (f1a2970fd)  median     8.632 ms/step  [min     8.422, max     8.757]  gerrychain 1.0.0a2    (1.43x baseline speed)
local (working tree)                                      median     6.009 ms/step  [min     5.856, max     6.081]  gerrychain 1.0.0a2    (2.06x baseline speed)

==== summary: graph=grid_50x50.json, 5 seed(s) x 1 repeat(s), 1000 steps, parts=25, epsilon=0.01, jobs=5 ====
pypi:0.3.2                                                median     9.626 ms/step  [min     9.301, max    10.306]  gerrychain 0.3.2      (baseline)
git:f1a2970fd5df23470a560eba440db2a9e8c624ba (f1a2970fd)  median     6.771 ms/step  [min     6.287, max     7.206]  gerrychain 1.0.0a2    (1.42x baseline speed)
local (working tree)                                      median     4.737 ms/step  [min     4.387, max     5.074]  gerrychain 1.0.0a2    (2.03x baseline speed)
```

## Why

After some user testing, we identified there were performance bottlenecks in the hotpath that reduced the expected performance uplift from the rustworkx migration (we were targeting ~2×, but some pathological examples showed very little improvement). This PR addresses some of the improvement areas that we identified.

Bottlenecks were identified via a combination of user feedback and profiling with the new `benchmarks/benchmark_recom.py` script included in this PR.

```bash
uv run python benchmarks/benchmark_recom.py run --profile --num-seeds 1
```

Profiling chains after the rustworkx migration showed most of the remaining overhead was wrapper bookkeeping rather than graph algorithms: `__getattribute__` interception alone produced tens of millions of redundant calls per chain, every proposal paid for full graph validation, and contiguity checks built full `Graph` subgraph wrappers (with node-id translation maps) just to answer yes/no connectivity questions. This PR clears that out. With these changes, ReCom on the hard graph runs ~1.5–1.8× faster than the 0.3.2 release, confirming the migration is a net performance win.

## Changes

- **`FrozenGraph` delegation**: replaced the `__getattribute__` override (which intercepted every attribute access and re-entered itself) with `__getattr__`, so attributes defined on `FrozenGraph` resolve via the normal fast path. Dunder probes (`__deepcopy__`, etc.) are guarded so copy/pickle/introspection no longer leak into the embedded graph; index caches moved into slots.
- **Validation moved to the construction boundary** with a maintainer-facing switch (`gerrychain.set_runtime_checks` / `runtime_checks()` context manager, modeled on `numpy.seterr`): cheap structural invariants stay always-on, while O(nodes + edges) audits are opt-in (off by default, enabled throughout the test suite). Per-call `verify_graph_is_valid` is gone from the hot path.
- **Graphs are forced undirected at construction**, resolving the `is_directed()` concern from migration review.
- **Spanning-tree fast path** for the non-region-aware case, skipping region-surcharge work when no region weights are in play.
- **Contiguity constraint pass**: `_are_reachable` rewritten as a plain deque BFS (heap/counter/distance dicts removed; per-iteration target rescan replaced with a shrinking unseen-set); the per-edge avoid closure inlined to a single dict lookup; new `Graph.is_node_set_connected()` runs the backend-native connectivity check directly instead of building a `Graph` subgraph wrapper; `Graph.neighbors()` no longer copies rustworkx's `NodeIndices` into a list; dead code deleted (`_bfs`, obsolete `contiguous_bfs` stub).
- **Test suite updates**: new grid fixtures in `conftest.py`, runtime checks enabled suite-wide, region-aware tests reworked, obsolete frm-graph tests removed.

## Testing

- [x] Tests pass locally (`PYTHONHASHSEED=0 uv run pytest --runslow` - 265 passed, 2 xfailed)
- [x] Added/updated tests (conftest fixtures, region-aware rework, `is_directed` coverage)
- [x] Manually tested relevant behavior (benchmarks shown above)

ReCom benchmarked on several graphs (PA, hard_graph, grid_8x8, grid_25x25, grid_50x50) with 5 seeds each, showing a consistent ∼20-30% speedup over the previous merge (f1a2970fd5df23470a560eba440db2a9e8c624ba) and an ∼2× speedup over the 0.3.2 release.

## Reviewer Notes

- **`Graph.neighbors()` return type changed** from `list` to a re-iterable sequence (`rustworkx.NodeIndices` on the RX path; the NX path still returns a list since NX hands back a single-pass iterator). All in-repo callers just iterate, but external code using list methods on the result would break — worth a changelog entry.
- **Thorough validation is now opt-in.** Anyone relying on implicit per-call validation should use `gerrychain.set_runtime_checks(True)` or the `runtime_checks()` context manager while debugging. The `random.choice` in `single_flip_contiguous` is deliberately kept even though any start node works: removing it would shift the global RNG stream and change every seeded chain's flip sequence (commented in the code).
- `contiguous()` no longer warms the `partition.subgraphs` cache; nothing in the repo relied on the constraint having populated it.
- `_are_reachable`'s signature changed (avoid-callable → mapping + part), but it is module-private. `contiguous_bfs` was removed outright rather than deprecated — it was unexported and only ever raised.
- Riskiest area for review: the `FrozenGraph.__getattr__`/slots rework and the construction-boundary validation, since both touch object-model behavior rather than just speed.


gerrychain/graph/graph.py
```diff
@@ -1191,42 +1245,46 @@ def __len__(self) -> int:
         return len(self.node_indices)
 
     def __getattr__(self, __name: str) -> Any:
-        """If the user requests the value of an attribute that is not defined for a GerryChain.
+        """Delegate unknown attributes to the embedded backing graph.
+
+        Accesses that Graph itself does not define are forwarded to the embedded NetworkX or
+        RustworkX graph. This keeps backend conveniences reachable - e.g. ``graph.geometry``
+        (stored on the NX graph) or ``graph.graph["crs"]`` (the NX graph-level attribute dict).
+
```
chief-dweeb: Is there any example you can dream up for when it would make sense to delegate to the embedded NX or RX graph?  In order for that to make sense, the caller would either have to already know whether the embedded graph was NX or RX, or else both NX and RX would need to have the exact same name (and semantics) for the delegated function call.

This seems dangerous to me...  Or if not dangerous then kind of not really useful...

gerrychain/graph/graph.py
```diff
@@ -1833,22 +1888,20 @@ def node_data(self, node_id: Any) -> dict[Any, Any]:
...
             )
 
-        if not isinstance(data_dict, dict):
-            raise TypeError("graph.node_data(): data for node is not a dict")
-
-        return data_dict
+        # if not isinstance(data_dict, dict):
+        #     raise TypeError("graph.node_data(): data for node is not a dict")
+        #
+        # return data_dict
```
chief-dweeb: Why leave this code in place but commented out - why not just delete the code?

gerrychain/graph/graph.py
```diff
@@ -2020,26 +2071,65 @@ def create_scipy_sparse_array_from_rx_graph(
...
-    # This code replaced calls on nx.is_connected()
-    def is_connected_bfs(self):
-        """
-        Checks if an undirected graph is connected using BFS.
+    def is_connected(self):
+        """Return whether the (undirected) graph is connected.
+
+        Delegates to the backend's native connectivity routine - ``rustworkx.is_connected`` for an
+        RX graph, ``networkx.is_connected`` for an NX graph - which are faster than hand-rolled
+        Python traversal.
```
chief-dweeb, line 2079: Nice

gerrychain/graph/graph.py
```diff
@@ -2020,26 +2071,65 @@ def create_scipy_sparse_array_from_rx_graph(
...
+            )
+
+    def is_node_set_connected(self, nodes: Iterable[Any]) -> bool:
+        """Return whether the given set of nodes induces a connected subgraph of this graph.
 
-        node_ids = list(self.node_indices)
+        This is a fast path for connectivity checks. It hands the node set straight to the
+        backend's subgraph constructor and runs the native connectivity routine on the result,
+        skipping the ``Graph`` wrapper and the node-id translation maps that ``subgraph()``
+        builds - none of which are needed to answer a yes/no connectivity question.
```
chief-dweeb, line 2108: Again - nice!

gerrychain/graph/graph.py
```diff
@@ -2283,7 +2373,10 @@ class FrozenGraph:
     # that the graph will not be modified.
     #
 
-    __slots__ = ["graph", "size"]
+    # Note: __slots__ means FrozenGraph instances have no __dict__, so we cannot use
+    # functools.cached_property (which caches into the instance __dict__). The cached
+    # node_indices / edge_indices values are stored in dedicated slots instead.
+    __slots__ = ["graph", "size", "_node_indices", "_edge_indices"]
```
chief-dweeb, line 2379: Interesting - I did not realize that __slots__ interacted with functools.cached_property.  

Your solution looks fine to me.

Out of curiosity, do you think cacheing the node and edge indices actually results in significantly better performance?

gerrychain/graph/graph.py
peterrrock2: Thank you!

gerrychain/graph/graph.py
peterrrock2: Thank you again!

gerrychain/graph/graph.py
peterrrock2: Because I forgot to.... fixing now

gerrychain/graph/graph.py
peterrrock2: Yeah, you are probably right here. Right now, it should be fine since every shared-name method is explicitly defined on Graph so it never falls through to `__getattr__`, but that's pretty fragile. 

I think that the correct thing to do here is to only allow the forwarding for the NX backend. 

The reason for keeping the NX-delegation is twofold: 
1. we need access to the `.geometry` attribute set when we create a graph from a geodataframe.
2. since the old version of the `Graph` class was a subclass of NetworkX's graph class, so it's generally expected by users that they can pass a GerryChain graph to NetworkX's algorithms and everything will work out fine. 

The RX backend doesn't really buy us anything, however, so it's best to just get rid of it and raise an error here.

gerrychain/graph/graph.py
peterrrock2: Well, on a vanilla chain with only incremental updaters (e.g. population) it doesn't do that much. But if you use the `region_splits` updater or something that requires you to walk the nodes every time, the savings can significant enough to matter.

This was mostly a hygiene thing: the `FrozenGraph` class is supposed to be immutable, so the `node_indices` and `edge_indices` need to be invariant.

That being said, I realize now that I return a `set` rather than a `frozenset`, so that needs to be fixed.

gerrychain/graph/graph.py
chief-dweeb: I agree - delegate in NX case but throw an error in the RX case...

chief-dweeb:
I want to spend some time better understanding the rather impressive work you did to compare performance, but that should not stand in the way of closing this PR and merging everything.

I am assuming that you will close this PR even though you created it.

PR #445, opened 2026-06-11
Opt/incorporate jeanne bfs ideas

peterrrock2:
## Summary

Post-migration performance pass to be reviewed following #444. This PR makes some simple modifications to the BFS implementation that we use to find balanced edge cuts. The main savings come from come from combining the computation of predecessors and successors into a single pass function.

Improvements here seem to show a ~20-30% speedup over the previous merge (3785ab1aeffe3b352e31941b65929d745c5a1dd6). Speed improvements were measured using the new benchmark suite included in this PR.

```bash
uv run python benchmarks/benchmark_recom.py \
  compare pypi:0.3.2 git:3785ab1aeffe3b352e31941b65929d745c5a1dd6 local \
  --graph benchmarks/graphs/<GRAPH> \
  --parts <PARTS> \
  --steps <STEPS>
```

### Results on some graphs

**PA.json: 1000 steps, 17 parts, 5 seeds, 1 trial per seed, ε=0.01**

| GerryChain Version  | Min Observed Time per Step | Max Observed Time per Step | Median Time per Step | Speedup vs. 0.3.2 |
| ------------------- | -------------------------- | -------------------------- | -------------------- | ----------------- |
| 0.3.2 (PyPI)        | 93.991 ms                  | 133.540 ms                 | 109.368 ms           | Baseline          |
| 1.0.0a2 (3785ab1ae) | 41.105 ms                  | 44.178 ms                  | 42.802 ms            | 2.56×             |
| 1.0.0a2 (local)     | 34.510 ms                  | 37.147 ms                  | 35.501 ms            | 3.08×             |

**hard_graph.json: 250 steps, 6 parts, 5 seeds, 1 trial per seed, ε=0.01**

| GerryChain Version  | Min Observed Time per Step | Max Observed Time per Step | Median Time per Step | Speedup vs. 0.3.2 |
| ------------------- | -------------------------- | -------------------------- | -------------------- | ----------------- |
| 0.3.2 (PyPI)        | 826.499 ms                 | 1327.474 ms                | 1053.129 ms          | Baseline          |
| 1.0.0a2 (3785ab1ae) | 460.001 ms                 | 685.946 ms                 | 553.602 ms           | 1.90×             |
| 1.0.0a2 (local)     | 381.407 ms                 | 543.359 ms                 | 449.221 ms           | 2.34×             |

**grid_8x8.json: 10000 steps, 8 parts, 5 seeds, 1 trial per seed, ε=0.01**

[...]

PR #447, opened 2026-06-11
Fix non-symmetric matrix bug in new normalized laplacian method

peterrrock2:
## Summary

To be reviewed following #445.

Fixes a correctness bug in `Graph.normalized_laplacian_matrix()` on the RustworkX code path. The matrix was built from a one-directional adjacency matrix (each undirected edge recorded as `A[u, v]` but never `A[v, u]`), so it came out **asymmetric** with the wrong node degrees and wrong normalization. The NetworkX path was unaffected.

This matters in practice because `Partition` converts graphs to RustworkX, so every real partition uses the RX path. `spectral_recom` takes the eigenvectors of this matrix to pick a cut, so it was operating on the wrong operator.

## Why

`normalized_laplacian_matrix()` builds the adjacency matrix `A` from `rx_graph.edge_list()`, which yields each undirected edge exactly once. The old code only appended the `A[u, v]` entry and never its mirror `A[v, u]`, leaving `A` as a single triangle of the true (symmetric) adjacency matrix. Everything downstream then goes wrong:

- `diags = A.sum(axis=1)` (the node degrees used to normalize) is incorrect - a node that only ever appears as the _second_ endpoint of its edges is assigned degree 0.
- The returned `D^{-1/2} (D - A) D^{-1/2}` is asymmetric and numerically wrong.

`proposals/spectral_proposals.py::spectral_recom()` feeds the result straight into `numpy.linalg.eigh`, which assumes a symmetric/Hermitian matrix and only reads one triangle. The
in-code comment there even asserts "we have a symmetric matrix" - which was not true on the RX path. The vector used to bipartition the region was therefore computed from the wrong matrix.

Concretely, for the graph `[(0, 1), (0, 2), (1, 2), (2, 3)]`:

```console
reference (networkx.normalized_laplacian_matrix):
 [[ 1.    -0.5   -0.408  0.   ]
  [-0.5    1.    -0.408  0.   ]
  [-0.408 -0.408  1.    -0.577]
  [ 0.     0.    -0.577  1.   ]]

RX path (before this fix):
 [[ 1.    -0.707 -0.707  0.   ]
  [ 0.     1.    -1.     0.   ]
  [ 0.     0.     1.     0.   ]
  [ 0.     0.     0.     0.   ]]   # asymmetric; node 3 dropped to degree 0
```

## Changes

- `gerrychain/graph/graph.py`: in the RX branch of `normalized_laplacian_matrix()`, add the mirrored entry `A[v, u]` for every edge `A[u, v]` (skipping self-loops so the diagonal is not double counted). This makes the adjacency matrix symmetric, which fixes both the degrees and the normalization.
- Removed the stale `frm: TODO: Get someone to validate ...` comment; the result is now validated against `networkx.normalized_laplacian_matrix()` in the test suite.

## Testing

- [x] New tests pass locally (`tests/test_laplacian.py`, `tests/test_proposals.py`,
      `tests/test_frm_graph.py` - 39 passed, including the `spectral_recom` proposal test)
- [x] Full test suite passes locally (`make test-all` - 265 passed, 2 xfailed)

New tests in `tests/test_laplacian.py`:

- RX normalized Laplacian matches the `networkx.normalized_laplacian_matrix()` reference over a set of randomized **connected** graphs (sizes 2-64, several seeds each), comparing both graphs over the same node ordering. Each case also asserts the result is symmetric.
- A dedicated symmetry regression test (direct guard for this bug).
- NX-based and RX-based `Graph` objects produce the same normalized Laplacian.
- An isolated-node (degree 0) graph, which exercises the divide-by-zero guard (inf -> 0) and checks for an all-zero row/column rather than NaNs.

Confirmed the new tests fail on the pre-fix code and pass after the fix (the pre-existing non-normalized `laplacian_matrix()` test passes in both, since that path was never broken).

## Reviewer Notes

- The non-normalized `laplacian_matrix()` was already correct - it uses `rustworkx.adjacency_matrix()`, which returns the full symmetric matrix. Only the hand-rolled
  normalized path was affected.
- Self-loops are added once on purpose. GerryChain graphs are simple, but this keeps the diagonal correct if a self-loop ever appears.

### DIFF
--- a/gerrychain/graph/graph.py
+++ b/gerrychain/graph/graph.py
@@ -2009,6 +2009,14 @@ class Graph:
                 cols.append(v)
                 data.append(1)  # simple adjacency matrix, so just 1 not weight attribute
 
+                # rx_graph.edge_list() yields each undirected edge only once, but the
+                # adjacency matrix of an undirected graph is symmetric, so we must add the
+                # mirrored entry as well. Skip self-loops so the diagonal is not double counted.
+                if u != v:
+                    rows.append(v)
+                    cols.append(u)
+                    data.append(1)
+
             sparse_array = scipy.sparse.coo_matrix(
                 (data, (rows, cols)), shape=(num_nodes, num_nodes)
             )
@@ -2040,12 +2048,9 @@ class Graph:
 
             """
 
-            # frm: * TODO:  Get someone to validate that this in fact does the right thing.
-            #
-            # The one test, test_proposal_returns_a_partition[spectral_recom], in test_proposals.py
-            # that uses normalized_laplacian_matrix() now passes, but it is for a small 6x6 graph
-            # and hence is not a real world test...
-            #
+            # The RX result is validated against networkx.normalized_laplacian_matrix()
+            # (the reference implementation used by the NX path) in
+            # tests/test_laplacian.py.
 
             A = create_scipy_sparse_array_from_rx_graph(rx_graph)
             n, _ = A.shape  # shape() => dimensions of the array (rows, cols), so n = num_rows
--- a/tests/test_laplacian.py
+++ b/tests/test_laplacian.py
@@ -1,3 +1,5 @@
+import random
+
 import networkx as nx
 import numpy as np
 import pytest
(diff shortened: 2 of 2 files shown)


chief-dweeb:
I am really glad that you found this.  The code I created for the normalized laplacian was just lifted from the web, and I knew that the test suite was not adequately testing it.  Thanks for taking a look and catching the bug!

OK to close and merge from my POV...

peterrrock2:
Great! Thank you for flagging this for review!

PR #448, opened 2026-06-12
Standardize wording and module doc strings for 1.0.0  (pass 1)

peterrrock2:
## Summary

To be reviewed following #447.

This PR standardizes a number of inconsistent public names, adds one new spanning-tree option and a bundled example graph, removes
one dead updater, and fills in a lot of missing documentation.

There are **no algorithmic or performance changes** here - ReCom produces the same plans as before, with two intentional exceptions: the `flips` updater was removed, and `random_spanning_tree` gained a new (default-off) behavior flag.

## Why

Code review on the RustworkX migration surfaced a batch of naming inconsistencies and documentation
gaps:

- `reg` vs `region`, `build_propose_<x>_proposal` (a "propose"/"proposal" stutter), and `L1_` vs `L_minus_1_` were all inconsistent.
- A `flips` updater existed that just returned `partition.flips` (flagged in-code as "kind of useless").
- Core pieces (`Partition`, `updaters/flows`, `Graph.from_json`/`to_json`, `random_spanning_tree`) lacked the documentation needed to understand how they work.
- Region-aware examples had to load `gerrymandria.json` from a relative path, which is duplicated across the repo.

### Breaking changes

- **Proposal builders renamed `build_<x>_proposal` -> `build_<x>_proposal_fn`** (all 9: `build_recom_proposal_fn`, `build_reversible_recom_proposal_fn`, `build_spectral_recom_proposal_fn`, `build_any_node_flip_proposal_fn`, `build_flip_every_district_proposal_fn`, `build_chunk_flip_proposal_fn`, `build_random_flip_proposal_fn`, `build_slow_reversible_proposal_fn`, `build_slow_reversible_bi_proposal_fn`). Not really breaking since these are new, but wanted to flag.
- **L-norm constraints renamed `L1_... -> L_1_...`** (`L_1_polsby_popper`, `L_1_reciprocal_polsby_popper`) so the `L_1_*` names line up with the existing `L_minus_1_*` family.
- **Removed the `flips` updater** (`gerrychain.updaters.flips`). It only returned `partition.flips`, which is easily accessed directly.
- **Region-split internals renamed to spell out `region`**: `total_reg_splits -> total_region_splits`, and `reg_attr`/`reg_attr_lst` parameters -> `region_attr`/`region_attr_lst` in `updaters/county_splits.py`. The public `tally_region_splits` is unchanged. ("region" is now the generic admin-unit term; county/locality remain specific instances.)

### New features

- **`random_spanning_tree(..., treat_unassigned_as_single_region=False)`**: selects how nodes with no value for a surcharged attribute (region-less / `None`) are treated. `False` (default) keeps the previous behavior (region-less nodes are individually splittable); `True` treats all region-less nodes as one additional region to keep whole.
- **`bipartition_tree(..., spanning_tree_fn_kwargs=None)`**: a dict forwarded verbatim to the spanning-tree function, so spanning-tree-specific options (such as `random_spanning_tree`'s `treat_unassigned_as_single_region`, or options of a custom spanning-tree function) can be set without naming each one. It lives only on `bipartition_tree` - the layer that actually calls the spanning-tree fn. `recom`/`build_recom_proposal_fn` deliberately do **not** carry it; you reach it by passing a pre-bound `bipartition_tree_fn` (the same idiom already used for `max_attempts`), e.g. `bipartition_tree_fn=partial(bipartition_tree, spanning_tree_fn_kwargs={"treat_unassigned_as_single_region": True})`. (`region_surcharge` remains the one spanning-tree option lifted to `recom` as a convenience, since it is the common case.) A new `SpanningTreeFn` `Protocol` (in `tree/bipartition_tree.py`, alongside `FindBalancedEdgeCutsFn`) documents the shared `(graph, region_surcharge=None, **kwargs) -> Graph` contract. This is intentionally a **super-user** entry point: not discoverable from the proposal signature and not type-checked per option.
- **`gerrychain.examples.gerrymandria()`**: the canonical 64-node region-aware "toy state" graph, now shipped as package data and loaded via `importlib.resources`, so it works from any directory (no file path needed). The duplicate `gerrymandria.json` under `tests/frm_tests/` was removed and those tests now use the constructor.

### Documentation

- New module docstrings for
  - `partition/partition.py` (assignment vs. the district view, FrozenGraph, lazy/incremental updaters, in- vs. out-of-chain use), and
  - `updaters/flows.py` (the incremental flow-updater pattern).

- Expanded function/method docstrings:
  - `Graph.from_json` and `Graph.to_json` (the on-disk adjacency format, NX-only constraint, geometry handling),
  - `random_spanning_tree` (a full "region surcharge" explanation),
  - `on_flow`/`on_edge_flow` ("initialize" at the root vs. incremental updates),
  - `nodes`/`node_indices` and `edges`/`edge_indices` (set vs. list; object vs. id), and
  - `Partition.keys`/`__getitem__` (the `[]` interface is over updaters, not nodes).

## Testing

- [x] `make test` passes: **292 passed, 2 xfailed**. (Two `test_data_and_geometry` tests fail in this environment due to a local shapely/geometry version mismatch; they are unrelated to these changes.)
- New tests: `tests/test_examples.py` (gerrymandria loads, has the expected region attributes, and loads from any working directory), and `test_spanning_tree_fn_kwargs_forwarded_to_spanning_tree_fn` in `tests/test_region_aware.py` (`bipartition_tree` forwards `spanning_tree_fn_kwargs` to the spanning-tree function, including through `recom` via a pre-bound `bipartition_tree_fn`; it defaults to no extra options, and `uniform_spanning_tree` is unaffected).
- Existing region-aware / tree / proposal / constraint / updater suites were updated for the renames.

## Reviewer Notes

- The proposal-builder and L-norm renames are deliberately **hard breaks** (no aliases) since we are
  pre-major-release.
- `spanning_tree_fn_kwargs` lives on `bipartition_tree` (the layer that actually calls the spanning-tree fn), **not** on `recom`/`build_recom_proposal_fn`. I am trying to lift only common options to `recom` as named params (`region_surcharge`), and treat remaining options as super-user only.
- Because `region_surcharge` is always bound through its own parameter, also supplying it _inside_ `spanning_tree_fn_kwargs` would otherwise raise an opaque `functools.partial() got multiple values for keyword argument 'region_surcharge'`. `_internal_bipartition_tree` now guards that case and raises a `ValueError` pointing at the fix ("Pass region_surcharge via the region_surcharge parameter, not inside spanning_tree_fn_kwargs."). Covered by `test_region_surcharge_inside_spanning_tree_fn_kwargs_raises`.


gerrychain/examples/__init__.py
```diff
@@ -0,0 +1,47 @@
+"""Small, self-contained example graphs that ship with GerryChain.
+
+These exist so that documentation examples, quick experiments, and tests can use a real graph
+without depending on an external data file or a particular working directory. The data is shipped
+inside the installed package and loaded via :mod:`importlib.resources`.
```
chief-dweeb, line 5: Thank you!  I had wanted to do just this, but I didn't know how to do so...

gerrychain/graph/graph.py
```diff
@@ -919,10 +1030,16 @@ def node_indices(self) -> set[Any]:
...
     @property
     def edge_indices(self) -> set[Any]:
-        """Return a set of the edge_ids in the graph.
+        """Return a ``set`` of the edge *ids* in the graph.
+
+        Unlike :meth:`nodes`/:meth:`node_indices` (which carry the same content up to a
+        permutation), ``edge_indices`` and :meth:`edges` are genuinely different: ``edge_indices``
+        returns edge *ids* (opaque integers under RustworkX), while :meth:`edges` returns the edges
+        themselves as ``(u, v)`` tuples of node_ids. Use an edge id with
+        :meth:`get_edge_from_edge_id` / :meth:`get_edge_id_from_edge` to convert between the two.
```
chief-dweeb, line 1039: Perfect - concise but complete (and understandable)

gerrychain/graph/graph.py
```diff
@@ -1016,29 +1133,30 @@ def get_edge_id_from_edge(self, edge: tuple[Any, Any]) -> Any:
...
+        That object-vs-id distinction is load-bearing for edges, but for nodes the node and its id
+        coincide, which is why ``nodes`` and ``node_indices`` carry the same content.
+
+        There is also a minor subtlety that users are unlikely to encounter unless accessing the
+        graph attribute off of a Partition object, but it is worth noting: the node_ids in the
+        graph attribute of a Partition object are not necessarily the same as the node_ids in the
+        it was common to create nodes whose ids were essentially names (the id carried semantic
+        original graph that was used to create the Partition object. Before the move to RustworkX
+        weight). That is not true of RustworkX node_ids, so any code that relies on the semantics
+        of a node's id (treating it like a name) is suspect in the RustworkX world.
```
chief-dweeb: Looks like some text got scrambled here.

"... same as the node_ids in the it was common to create..."

"...(the id carried semantic original graph that..."

"...Before the move to RustworkX weight)\..."

Also - while I get your point that NX node-ids might have semantic weight, it is also true that even if the NX node-ids did NOT have semantic weight, the RX node-ids will be different than the NX node-ids, so users should be very careful post analysis that they know what node-ids they are working with.

gerrychain/graph/graph.py
```diff
@@ -1060,10 +1178,14 @@ def nodes(self) -> list[Any]:
 
     @property
     def edges(self) -> set[tuple[Any, Any]]:
-        """Return a set of all of the edges in the graph, where each edge is a tuple of node_ids.
+        """Return a ``set`` of all of the edges in the graph, where each edge is a ``(u, v)`` tuple
+        of node_ids.
+
+        This returns the edges themselves, which may not be the same as their ids. For the edge
```
chief-dweeb: In RX the edges will ALWAYS be different from the edge-ids (integers vs. tuples)

gerrychain/partition/partition.py
```diff
@@ -1,3 +1,63 @@
...
+
+    * ``assignment`` (an :class:`~gerrychain.partition.assignment.Assignment`) is the
+      node -> part map. It is the source of truth and is cheap to update flip-by-flip.
+    * ``parts`` is the inverse, part -> set-of-nodes, view. Many updaters and the recombination
+      proposals need to iterate the nodes of a part, for which the inverse view is far more
+      convenient and efficient than scanning the entire assignment.
+
+    Keeping both means neither direction of lookup has to be recomputed from the other on every use;
+    the ``Assignment`` object keeps the two consistent as flips are applied.
+
```
chief-dweeb, line 25: This might be a place to talk about flows as well - that we keep track of the changes in assignments (and parts) via flows.  Perhaps a short summary and a pointer to better documentation elsewhere?

gerrychain/partition/partition.py
```diff
@@ -1,3 +1,63 @@
...
+    using the node/edge "flows" between parent and child. See :mod:`gerrychain.updaters.flows`.
+
+Inside vs. outside a Markov chain:
+    A Partition is most useful as chain state, but it is also handy for post-hoc analysis of a fixed
+    plan (computing its cut edges, population deviation, partisan metrics, etc.). The distinguishing
+    feature is the ``parent``: a chain produces a lineage of partitions, each with a parent, which
+    is exactly what the incremental/flow updaters exploit. A standalone Partition built directly
+    from a graph and an assignment has ``parent is None``; its flow-based updaters then simply fall
+    back to computing everything from scratch (their "initializer" path). The same updaters
+    therefore work in both settings.
```
chief-dweeb, line 60: I don't know if it is worth it to mention that the node-ids in assignments in a chain will be RX node-ids, so any code that looks at post chain node-ids should be careful - it might need to translate those node-ids back to the original NX node-ids (there are functions to do this translation)

gerrychain/partition/partition.py
```diff
@@ -1,3 +1,63 @@
+"""The :class:`Partition` - an assignment of graph nodes to districts, plus cached updaters.
```
chief-dweeb, line 1: Thank you for writing up what Partitions are for - excellent!

gerrychain/partition/partition.py
```diff
@@ -379,6 +441,20 @@ def __getattr__(self, key: str) -> object:
...
+        returns the updater names (e.g. ``["cut_edges", "population", ...]``) - not node_ids or
+        part_ids. This often surprises people who think of a Partition as a node-to-district map.
+
+        The actual node-to-part data lives in :attr:`assignment` (``partition.assignment``, a
+        node_id -> part_id mapping), and the inverse part-to-nodes view is :attr:`parts`
+        (``partition.parts``).
+
+        Returns:
+            KeysView[str]: A view of the updater names available on this partition.
+        """
```
chief-dweeb, line 473: Nice

gerrychain/tree/spanning_tree.py
```diff
@@ -90,40 +90,99 @@
...
+        random weight when the edge crosses a boundary for that attribute - that is, when the two
+        endpoints do NOT share the same non-null value of the attribute. So edges *inside* a region
+        keep their plain random weight, while edges on the *boundary* of a region are made heavier.
+
+        Since this function builds a *minimum* spanning tree, making the boundary edges heavier
+        biases the MST toward connecting each region through its cheap interior edges. Therefore,
+        the region tends to appear as a single connected subtree of the spanning tree, and
+        because the bipartition step cuts exactly one edge, a connected region tends to be kept
+        whole. In the extreme case, if the surcharge is large enough to dominate the random weights,
+        then the region will be split at most once.
```
chief-dweeb: I am missing something here - what do you mean when you say that a region will be split at most once?  If a connected region has a population that is larger than two districts, then won't it be cut twice?

Stated differently, I think your comment assumes that the population of a region is less than the target population of a district...

But perhaps I am missing something...

gerrychain/tree/spanning_tree.py
```diff
@@ -90,40 +90,99 @@
...
+          gets no "keep whole" bias and may be divided freely among districts. This is usually the
+          right behavior when region-less nodes are scattered (water, unincorporated areas, etc.),
+          where forcing them to stay together would make little sense.
+        * **Unassigned area kept whole** (``treat_unassigned_as_single_region=True``): edges between
+          two region-less nodes are left cheap (since they share the value ``None``), so all
+          region-less nodes are treated like one additional region and biased to stay whole.
+
+        The flag is applied per attribute, so for a given attribute its region-less nodes are
+        treated consistently. It has no effect unless some nodes actually lack a value for a
+        surcharged attribute.
```
chief-dweeb, line 158: Is there a nice example of when this behavior would be desirable?

gerrychain/updaters/flows.py
```diff
@@ -1,3 +1,50 @@
...
+  no parent to diff against.
+* an **update rule** (the decorated function), which is called once per changed part on every
+  later step and returns the new value for that part, given the parent's value for that part plus
+  the in/out flow for that part.
+
+So the lifecycle of a flow-based updater is: compute everything once at the root via the
+initializer, then on each subsequent step copy the parent's result and patch only the parts that
+actually changed. This is what the docstrings below mean by "initialize" (root) vs. the incremental
+"post-initialization" update (every later step).
+"""
```
chief-dweeb, line 46: Thank you, thank you, thank you...

I think that this will help anyone reading the code better understand what is going on.

Nicely done.

chief-dweeb:
Fine by me to close and merge - so much cleaner, so much easier to understand!

gerrychain/examples/__init__.py
peterrrock2: Absolutely!

gerrychain/graph/graph.py
peterrrock2: Ah, good note. I'll correct that

gerrychain/partition/partition.py
peterrrock2: Yeah, that wouldn't hurt. I'll add something short in.

gerrychain/partition/partition.py
peterrrock2: I prefer explicit to implicit, and I'm totally fine with longer module doc strings, so I'll add that in.

gerrychain/tree/spanning_tree.py
peterrrock2: Good catch! You are right. Classic example is LA County which must be split at least 12 times IIRC. I was thinking about sub-district sized regions when I wrote this.

gerrychain/tree/spanning_tree.py
peterrrock2: We have used both of these modes when thinking about splits for municipalities. Somewhat unintuitively, which version one "works better" in terms of preserving the municipalities (or both municipalities and counties) can depend on the geographic region of interest.

gerrychain/updaters/flows.py
peterrrock2: Abosolutely! Glad that you like it!

PR #450, opened 2026-07-11
Feat/better rng handling

peterrrock2:
## Summary

This PR addresses some of the major headaches that we have had around reproducibility for GerryChain. It replaces GerryChain's ambient random-number state with explicitly owned RNGs. Users can provide an integer seed or `random.Random` instance at the main ownership boundary, such as `MarkovChain`, without managing a package-global RNG.

It also closes a remaining reproducibility gap discovered after the RNG ownership rewrite: a seeded RNG only fixes the sequence of random values, not which graph elements receive those values. RNG- sensitive traversal is now deterministic for NetworkX-backed graphs, including initial plans made with `Partition.from_random_assignment()` and spanning trees on string-labeled nodes such as GEOIDs.

> NOTE: Despite the diff count, this change is actually pretty small. A chunk of the diff is from an SVG change, and then a lot of it is tweaking function calls.

## Why

The previous global RNG approach made independently seeded and interleaved chains difficult to reason about. It also forced users to set environment variables to make the chain reproducible, which is an anti-pattern that we wanted to address.

This change gives each chain or optimizer an independent RNG while preserving convenient integer- seed reproducibility.

The initial implementation still allowed `PYTHONHASHSEED` to affect seeded results. Some NetworkX paths consumed random values while iterating over sets of node or edge IDs. Python randomizes string hashes between processes, so the same random values could be assigned to different graph elements. That affected `Partition.from_random_assignment()`, random and uniform spanning trees, and therefore ReCom trajectories. Stable insertion-order traversal is required in addition to a seeded RNG.

### Library

- Add support for `rng=random.Random(...)`, `rng=<integer seed>`, and `rng=None` across chains, optimizers, initial-partition generators, proposals, and tree algorithms.
- Have `MarkovChain` own one RNG and pass it explicitly to proposals and acceptance functions through the keyword-only `rng` argument.
- Make stochastic tree and ReCom callbacks receive the owning RNG, including custom balanced-cut, spanning-tree, root-choice, and cut-choice functions.
- Preserve the original graph and node labels when `Partition.from_random_assignment()` calls its partitioning method. `Graph.nodes` now clearly provides backend insertion order, so initial-plan generation does not populate a temporary graph from unordered node indices.
- Give NetworkX-backed tree sampling stable RNG-sensitive traversal:
  - assign random edge weights in insertion order rather than set order;
  - choose and visit nodes in insertion order in `uniform_spanning_tree()`;
  - use insertion order when balanced-cut finders choose roots and collect leaves; and
  - leave the established RustworkX integer-index paths unchanged to preserve existing trajectories.
- Make seeded initial assignments, random spanning trees, uniform spanning trees, and ReCom trajectories independent of `PYTHONHASHSEED`, including graphs with string node labels.
- Change the public `node_repeats` default from 1 to 0 throughout ReCom, bipartitioning, and initial-partition generation. An unsuccessful exhaustive search now redraws the spanning tree immediately.
- Define `node_repeats` as the number of _additional_ roots tried on the same tree. Warn when a positive value is combined with the default memoized cut finder, which already examines every edge. Positive values remain useful with contraction or a custom root-sensitive cut finder.
- Resolve default random root and cut choosers from the supplied RNG instead of `random.choice`. Custom cut-choice functions use `cut_choice_fn(cuts, *, rng)` signature.
- Prevent partitions from mutating the class-level default updater dictionary.

### Docs

- Add a seeded asset generator and regenerate the GerryMandria animations and Pennsylvania plot. The ReCom guide now uses explicit seeds, deterministic examples, and consistent Districtr colors.
- Expand documentation for self-loops, surcharge strength, `node_repeats` versus `max_attempts`, pair reselection, custom cut choosers, chain mixing caveats, and reproducibility limits.
- Incorporate the @chief-dweeb's tutorial corrections and clean up API docstrings so Sphinx can build the updated documentation without warnings.

## Testing

- [x] Tests pass locally
- [x] Added/updated tests
- [x] Documentation builds with warnings treated as errors
- [x] ReCom benchmark shows no slowdown

Specific regression coverage includes:

- subprocess tests across multiple `PYTHONHASHSEED` values for string-labeled initial assignments, random spanning trees, uniform spanning trees, and chain trajectories;
- a custom `Partition.from_random_assignment()` method test confirming that it receives the original graph and its insertion-ordered node labels;
- tests that all public `node_repeats` defaults are 0 and that only memoized cut finders warn;
- seeded callback-contract and RNG-precedence tests; and
- an updated Pennsylvania trajectory hash captured after the deterministic-ordering correction.

## Reviewer Notes

- This intentionally changes the proposal and acceptance contracts. Custom callables must accept a keyword-only `rng` argument.
- Custom cut-choice functions use the signature `cut_choice_fn(cuts, *, rng)`. Custom spanning-tree and balanced-cut callbacks also receive `rng`.
- `MarkovChain` owns the RNG, and its call-time `rng` takes precedence over an RNG bound into a partially applied proposal.
- Reiterating the same chain resets its state but continues its RNG stream. Construct a new chain with the same integer seed to replay a trajectory.
- `Graph.node_indices` remains a set. RNG-sensitive NetworkX code must use `Graph.nodes` or another explicitly ordered traversal. `Partition.from_random_assignment()` passes its original graph to the selected method, which is important for retaining insertion order and non-integer node labels.
- The deterministic-ordering correction intentionally preserves the existing RustworkX traversal so established integer-index trajectories do not change unnecessarily.
- `node_repeats=0` is a behavioral default change. Code that explicitly sets a positive value still works, but it warns with the memoized cut finder because rerooting cannot reveal a cut that its exhaustive search missed.
- The documentation images are generated with fixed seeds by `docs/generate_recom_assets.py`; and should match what is in the RST files.


docs/generate_recom_assets.py
```diff
@@ -0,0 +1,267 @@
+"""Regenerate the seeded images used by ``docs/user/recom.rst``."""
```
chief-dweeb, line 1: Brilliant!!!

gerrychain/tree/bipartition_tree.py
```diff
@@ -886,16 +926,39 @@ def _internal_bipartition_tree(
...
+    #       is now memoization.
+    if node_repeats > 0 and unwrapped_cut_finder is find_balanced_edge_cuts_memoization:
+        warnings.warn(
+            "node_repeats is not beneficial with `find_balanced_edge_cuts_memoization`, which "
+            "exhaustively searches each spanning tree. Set node_repeats=0 to redraw the tree "
+            "after an unsuccessful search. Positive values are useful only with "
+            "`find_balanced_edge_cuts_contraction` or a custom cut-edge finder.",
+            UserWarning,
+            stacklevel=2,
+        )
```
chief-dweeb, line 948: This is very clever - kudos!

chief-dweeb (commented):
Wow - lots of changes.  I am impressed by the clarity of your updates to the RST files.  Thanks!

The RNG work was pervasive - it shows that you know the codebase quite well.  I am relieved to be told that what I thought of as fragility was in fact not of concern.

Go ahead and merge...

-Fred

gerrychain/tree/bipartition_tree.py
peterrrock2: Thank you for looking over this so thoroughly and raising your concerns! There is a lot of stuff that is hanging around in this codebase that I vaguely remember being issues, but my brain is split between a lot of projects making it easy to lose track of things, and it's incredibly helpful to have someone so detail oriented working on this with me and to pushing me as a maintainer to make this project better.

PR #454, opened 2026-07-13
Fix Gingleator empty list maximum

### LINKED ISSUE #424: max() over empty list in Gingleator
In Gingleator, if _all_ districts become majority-minority, then an error occurs on line 144 of reward_partial_dist(). In this case, it tries to take the max( ) over an empty list, and the code breaks. The same issue could arise on line 171 of reward_next_highest_close().

https://github.com/mggg/GerryChain/blob/a83acf55a371e42d2de0574e24569e2808b5e0d4/gerrychain/optimization/gingleator.py#L144

https://github.com/mggg/GerryChain/blob/a83acf55a371e42d2de0574e24569e2808b5e0d4/gerrychain/optimization/gingleator.py#L171

It would be great if GerryChain could terminate gracefully in such circumstances, either terminating the chain early because a global max has been found, or finishing the chain as intended but without returning an exception. Either seems reasonable to me.

campioru:
## Summary

This PR avoids `ValueError: max() iterable argument is empty` for globally optimal Gingle districts.

Closes #424.

## Why

As detailed in #424, the `reward_partial_dist()` and `reward_next_highest_close()` methods of `Gingleator` throw an error when all districts are majority-minority; they attempt to get the maximum of a list of below-threshold percentages, which is empty when all districts are above the Gingle threshold.

## Changes

- Set `next_dist = 0` if no districts are below the Gingle district threshold

## Testing

Added regression tests to avoid this in future.

- [x] Tests pass locally
- [x] Added/updated tests
- [ ] Not applicable

## Reviewer Notes

- No test (`pytest` and `flake8`) failed that did not also fail on main 4a66510.
- Regression tests have `threshold = 0.0` to test when all districts are majority-minority.

### DIFF
--- a/gerrychain/optimization/gingleator.py
+++ b/gerrychain/optimization/gingleator.py
@@ -138,7 +138,12 @@ class Gingleator(SingleMetricOptimizer):
         """
         dist_percs = part[minority_perc_col].values()
         num_opport_dists = sum(list(map(lambda v: v >= threshold, dist_percs)))
-        next_dist = max(i for i in dist_percs if i < threshold)
+
+        if num_opport_dists < len(dist_percs):
+            next_dist = max(i for i in dist_percs if i < threshold)
+        else:
+            next_dist = 0
+
         return num_opport_dists + next_dist
 
     @classmethod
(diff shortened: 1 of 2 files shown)


peterrrock2:
Hey @campioru thank you for working on this! Can you add a regression test or two so we don't ever run into this issue again?

campioru:
Thanks for the feedback @peterrrock2!

Regression tests added under `test_reward_<reward_type>` - let me know if these should be their own separate tests instead.

peterrrock2:
Looks good! Thanks for working on this!

PR #456, opened 2026-07-14
Update ReCom class to behave like a simpler namespace / interface for users

peterrrock2:
## Summary

This PR does not update any documentation since I would like feedback on the API design first.

This PR replaces the old callable `ReCom` class with a namespace of ready-made proposal builders for the standard ReCom variants.

It also adds explicit control over district-pair selection and fixes two RustworkX issues uncovered by builder-equivalence testing:

- RustworkX neighbor iteration was not deterministic.
- Uniform spanning trees could silently use node IDs that did not match the graph they spanned.

Together, these changes should make the common ReCom variants easier to discover and configure while preserving deterministic seeded trajectories.

## ReCom proposal namespace

`ReCom` is now a non-instantiable namespace with builders for the four combinations of pair selection and spanning-tree algorithm described in _Spanning Tree Methods for Sampling Graph Partitions_:

- `ReCom.cut_edges_mst`, alias `A`: cut-edge pair selection with a minimum spanning tree.
- `ReCom.district_pairs_mst`, alias `B`: district-pair selection with a minimum spanning tree.
- `ReCom.cut_edges_ust`, alias `C`: cut-edge pair selection with a uniform spanning tree.
- `ReCom.district_pairs_ust`, alias `D`: district-pair selection with a uniform spanning tree.
- `ReCom.reversible`, alias `R`: reversible ReCom.

Each builder returns a `ProposalFn` ready to pass to `MarkovChain`:

```python
from gerrychain.proposals import ReCom

proposal = ReCom.district_pairs_mst(
    pop_col="TOT_POP",
    pop_target=ideal_population,
    epsilon=0.01,
)
```

The named builders are intentionally more limited to reduce cognitive burden, but expose the most common options:

- `region_surcharge` for the minimum-spanning-tree variants;
- `allow_pair_reselection` for the four standard variants; and
- `max_balanced_edge_cuts` and `repeat_until_valid` for reversible ReCom.

Advanced users can continue to use `recom`, `build_recom_proposal_fn`, `reversible_recom`, or `build_reversible_recom_proposal_fn` directly.

## District-pair selection

`recom` and `build_recom_proposal_fn` now accept a `pair_selection` argument:

- `"district_pairs"` selects uniformly among adjacent district pairs and remains the default.
- `"cut_edges"` selects proportionally to the number of cut edges shared by each pair. This is equivalent to selecting a cut edge uniformly for the first attempt.

Candidate pairs are normalized and sorted before random selection so their order does not depend on Python hash iteration. Mixed-type district labels remain supported.

If bipartitioning fails and pair reselection is enabled, remaining candidate pairs are tried without replacement. Cut-edge-weighted candidates are drawn lazily, so a successful first attempt requires only one weighted draw.

Invalid `pair_selection` values now raise a clear `ValueError`.

### Deterministic neighbor order

RustworkX collects neighbors in a randomly seeded hash set, so its raw neighbor order can vary between otherwise identical calls.

`Graph.neighbors` now sorts RustworkX neighbors before returning them. This prevents seeded algorithms such as Wilson random walks and randomized breadth-first traversals from mapping the same random draws onto different neighbors.

### Preserved node IDs in uniform spanning trees

The RustworkX path in `uniform_spanning_tree` previously built a NetworkX tree and converted it back to RustworkX. That conversion could renumber nodes according to NetworkX insertion order.

Balanced-cut searches return subsets in the spanning tree's node-ID space. Applying a renumbered subset to the original graph could therefore split the wrong nodes.

Uniform spanning trees for RustworkX graphs are now constructed directly as RustworkX graphs, preserving the input graph's node IDs.

## Testing

- [x] Tests pass locally
- [x] Added/updated tests

Added coverage for:

- all named ReCom builders and their letter aliases;
- the non-instantiable namespace contract;
- balanced output from every standard variant;
- pair-reselection behavior;
- equivalence between each builder and its corresponding direct function call;
- reversible-builder equivalence;
- cut-edge weighting versus district-pair weighting;
- invalid pair-selection values;
- stable RustworkX neighbor order across repeated subgraph construction;
- repeatable uniform-spanning-tree ReCom trajectories; and
- the updated Pennsylvania reproducibility freeze.

Verified with

```
make test-all
```

Which shows

```
403 passed, 5 skipped, 2 xfailed,
```

### DIFF
--- a/gerrychain/graph/graph.py
+++ b/gerrychain/graph/graph.py
@@ -2036,9 +2036,12 @@ class Graph:
         is, all of the nodes that are directly connected to the given node by an edge.
 
         The result supports iteration (repeatedly), ``len()``, and indexing, but it is not
-        guaranteed to be a ``list``: for an RX graph it is the ``rustworkx.NodeIndices``
-        sequence returned by the backend, which avoids copying the neighbors into a fresh
-        list on every call. Callers that need list methods should wrap it in ``list()``.
+        guaranteed to be a ``list``. Callers that need list methods should wrap it in ``list()``.
+
+        The neighbors are returned in a deterministic order. RX collects neighbors into a randomly
+        seeded HashSet, so the raw ``rustworkx.NodeIndices`` order varies call to call; any seeded
+        algorithm that maps RNG draws over neighbor order (a random walk, a BFS feeding a random
+        choice) would otherwise be unreproducible.
 
         Args:
             node_id (Hashable): A node ID.
@@ -2047,7 +2050,7 @@ class Graph:
             Sequence[Hashable]: The neighboring node IDs.
         """
         if self._rx_graph is not None:
-            return self._rx_graph.neighbors(cast(int, node_id))
+            return sorted(self._rx_graph.neighbors(cast(int, node_id)))
         elif self._nx_graph is not None:
             # NX returns a single-pass iterator, so it must be materialized here;
             # callers (and the FrozenGraph.neighbors lru_cache) expect a re-iterable result.
--- a/gerrychain/proposals/tree_proposals.py
+++ b/gerrychain/proposals/tree_proposals.py
@@ -1,7 +1,7 @@
 import random
-from collections.abc import Callable, Hashable, Sequence, Set as AbstractSet
+from collections.abc import Callable, Hashable, Iterable, Iterator, Sequence, Set as AbstractSet
 from functools import partial
-from typing import cast
+from typing import Literal, cast
 
 from gerrychain._rng import make_rng
 from gerrychain.partition import Partition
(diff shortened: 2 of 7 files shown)


gerrychain/graph/graph.py
```diff
@@ -2036,9 +2036,12 @@ def neighbors(self, node_id: Hashable) -> Sequence[Hashable]:
...
         The result supports iteration (repeatedly), ``len()``, and indexing, but it is not
-        guaranteed to be a ``list``: for an RX graph it is the ``rustworkx.NodeIndices``
-        sequence returned by the backend, which avoids copying the neighbors into a fresh
-        list on every call. Callers that need list methods should wrap it in ``list()``.
+        guaranteed to be a ``list``. Callers that need list methods should wrap it in ``list()``.
+
+        The neighbors are returned in a deterministic order. RX collects neighbors into a randomly
+        seeded HashSet, so the raw ``rustworkx.NodeIndices`` order varies call to call; any seeded
+        algorithm that maps RNG draws over neighbor order (a random walk, a BFS feeding a random
+        choice) would otherwise be unreproducible.
```
chief-dweeb, line 2045: Good catch!

I wonder, however, if the additional sorting might cause a noticeable slowdown for large graphs with lots of steps in the chain.  One option would be to provide a parameter, do_not_sort_output, that defaulted to False to allow code that did not want to pay the performance penalty a way to avoid the sort.

However, if the performance penalty is not high, then simplicity wins (no additional parameter).

gerrychain/proposals/tree_proposals.py
```diff
@@ -461,251 +500,241 @@ def build_reversible_recom_proposal_fn(
...
-    ReCom (short for ReCombination) is a class that represents a ReCom proposal
-    for redistricting. It is used to create new partitions by recombining existing
-    districts while maintaining population balance.
+    """Ready-made builders for the standard ReCom proposal variants.
+
+    Each method returns a proposal function ready to hand to :class:`~gerrychain.MarkovChain`, so
+    instead of the ``functools.partial`` incantation::
+
+        from functools import partial
+        proposal = partial(recom, pop_col="TOTPOP", pop_target=ideal_pop, epsilon=0.01)
```
chief-dweeb, line 510: Stylistic nit - we seem to have opted for the suffix: "Fn" or "_fn" for function parameters / variables, so perhaps it should be proposal_fn = ... instead of just proposal = ...

Not a huge big deal - just a thought...

gerrychain/proposals/tree_proposals.py
```diff
@@ -461,251 +500,241 @@ def build_reversible_recom_proposal_fn(
...
+    spanning-tree distribution; see :func:`reversible_recom`.
+
+    The aliases ``A``, ``B``, ``C``, ``D``, and ``R`` name the same builders after the variants in
+    "Spanning Tree Methods for Sampling Graph Partitions" (Cannon et al., 2022,
+    https://arxiv.org/abs/2210.01401), to ease replication of results from that paper.
+
+    These builders expose only the most common options. For anything else (a custom bipartition or
+    spanning-tree function, ``node_repeats``, ...), use :func:`build_recom_proposal_fn` or
+    :func:`build_reversible_recom_proposal_fn`, which these methods wrap, or bind
+    ``functools.partial(recom, ...)`` yourself.
```
chief-dweeb, line 537: I really, really like this.  You leave legacy code working, but provide a much easier on-ramp for new users or for users who just want to do the straight-ahead stuff.

I would add additional verbiage, however, to help the user understand why there are so many different ways to do the same thing.  The point seems to me to be that expert users will want to be able to exert control over the operation of the algorithm, but that most users will not need that, so we wanted to provide a way to hide the complexity.  There are three levels: 1) use the namespace methods that package up almost everything, 2) use the build_xxx_proposal_fn() pattern if you want finer control, and 3) use functools.partial() if you grok what functools.partial() does and like that style.

While I would personally put this verbiage in the code - the most important place to put it is in the User Guide / Tutorial...

chief-dweeb (commented):
Really nice!

As I said in one of my comments - we will need to update the User Guide / Tutorial, but this is all good from my POV.

gerrychain/graph/graph.py
peterrrock2: Thanks! My paranoia tests caught this one by chance.

I checked this with some benchmarks, and it looks like the penalty is pretty negligible, and even if it wasn't, determinism is more important for this library.

gerrychain/proposals/tree_proposals.py
peterrrock2: Yes, agreed! I have that noted as a change for a different PR. I am trying to be better about resisting scope creep

gerrychain/proposals/tree_proposals.py
peterrrock2: Great! I am glad that you like it! Noted on the verbiage; I am always one for more documentation, so I will probably do both.

PR #458, opened 2026-07-19
Build a `MarkovChain` incrementally instead of all at once

peterrrock2:
## Summary

This PR lets users create a `MarkovChain` first and configure it piece by piece, rather than
assembling every argument before the chain can exist. Every constructor parameter is now optional,
and the chain can be completed by assigning attributes and calling small helper methods afterwards.

The old all-at-once construction still works unchanged; this only adds the incremental path.

## Incremental configuration

`MarkovChain.__init__` has been updated to make it so that most argruments are optional. This
allows legacy workflows to still work, but also enables a "Chain-first" paradigm:

```python
chain = MarkovChain(total_steps=1000, rng=2024)

chain.initial_partition = Partition(graph, assignment="2011_PLA_1")
chain.add_updaters({
    "population": Tally("TOT_POP", alias="population"),
    "cut_edges": cut_edges,
})
chain.proposal_fn = ReCom.district_pairs_mst(...)
chain.add_constraint(contiguous)
chain.acceptance_fn = always_accept
```

Assignment order does not matter. The full configuration is validated by `check_valid` when
iteration begins, so an incomplete or inconsistent chain is reported at the point it is run rather
than silently misbehaving. Attributes are also locked during iteration to prevent mid-run mutation.

## New helper methods

- `add_updater(name, updater)` / `add_updaters(mapping)`: register updaters on the chain. They are
  applied to the initial partition immediately if one is set, otherwise as soon as one is assigned,
  and every partition the chain produces inherits them.
- `add_constraint(constraint)` / `add_constraints(iterable)`: append to the chain's constraints.
  Each new constraint is checked against the initial partition as soon as it is added, so a starting
  plan that violates one is reported immediately.
- `check_valid()`: validate the complete configuration on demand.

While a run is in progress the chain is locked, and these mutators raise a clear error rather than
corrupting an in-flight iteration.

## Supporting changes

- A `ConstraintFn` type alias (`Callable[[Partition], bool]`) is introduced and used across the
  constraint and optimization APIs to slim down repeated signatures.
- The optimizers (`gingleator`, `optimization`) and `metagraph` are updated to the new chain
  surface.
- The tutorial notebooks are nudged toward the incremental style.

## Testing

- [x] Tests pass locally
- [x] Added/updated tests

`tests/test_chain.py` gains coverage for building a chain incrementally, adding updaters and
constraints before and after the initial partition is set, deferred validation at iteration time,
and the locking behavior that prevents mutation mid-run.

Verified with `make test`.


gerrychain/chain.py
```diff
@@ -49,179 +50,314 @@ class MarkovChain:
...
+        Raises:
+            AttributeError: If a run is in progress.
         """
-        self.counter = 0
-        self.state = self.initial_state
-        return self
+        if self.__locked:
+            raise AttributeError("Cannot add updater to a locked MarkovChain instance.")
+
+        self._added_updaters[name] = updater
```
chief-dweeb, line 211: When I read this code I had to go check to make sure that any _added_updaters that were added before an initial partition was set would be added later.  Might be worth a comment here that _added_updaters exists so that if no initial partition exists yet, the updaters can be added later on...

chief-dweeb:
This looks just fine to me.

Merge it!

gerrychain/chain.py
peterrrock2: Adding a comment never hurts!

Okay, I'm going to collect notes from this stack and put them into a new PR

PR #459, opened 2026-07-19
`_fn` naming convention and assignment-vector semantics

peterrrock2:
Mostly mechanical changes to the public API surface. Detailed review is not necessary. Don't worry about docs for now.

## Summary

Two independent pre-1.0.0 cleanups: every callable-valued parameter across the public API is renamed
to end in `_fn`, and `Partition` gains an assignment-vector view. Stacks on
`feat/builder-pattern-mc`.

## Why

Both are API-surface changes that are easier to make now than after 1.0.0 ships.

- Bare noun names for function parameters (`proposal`, `acceptance`, `beta`) do not read as
  callables at a call site, and were applied inconsistently across the codebase.
- Code that wants to hand a plan to array-oriented tooling had to rebuild the district assignment as
  a dense array by hand.

## Changes

- Renamed function-valued parameters to carry a `_fn` suffix (`proposal_fn`, `acceptance_fn`,
  `optimization_metric_fn`, `beta_fn`, `score_fn`, and so on) across the chain, acceptance
  functions, optimization, proposals, bounds, tally, flows, and the spanning-tree and bipartition
  APIs. Plural
  collection parameters such as `constraints` and `updaters` keep their noun names.
- Added a `Partition.assignment_vector` property returning the assignment as a `numpy.ndarray`
  indexed by node, with the supporting logic in `partition/assignment.py`.
- Updated the tutorial notebooks and `typing_assertions.py` to the renamed parameters.

## Testing

- [x] Tests pass locally
- [x] Added/updated tests

`tests/partition/test_assignment.py` and `tests/partition/test_partition.py` add coverage for the
assignment-vector API, and `tests/typing_assertions.py` pins the renamed public callable contracts.
Verified with `make test`.

## Reviewer Notes

- The `_fn` rename is a breaking change for any caller passing these parameters by keyword. It is
  purely a naming change; no behavior is affected.
- The rename is broad but mechanical, which is why it is isolated in its own PR rather than mixed
  into a behavioral change.


chief-dweeb:
Given your comments, I did not review the changes.  I assume that if all of the tests passed, you did not make any typos.

That said, if you would like me to review the changes, just let me know and I will do so.

PR #460, opened 2026-07-19
Make the test suite pass the type checkers

peterrrock2:
Mostly mechanical changes to the public API surface. No review necesasay. Don't worry about docs for now.

## Summary

Annotates and adjusts the tests so `ty` and `pyright` pass over `gerrychain` **and** `tests`, and
widens the type-check tooling to cover the tests. Almost entirely test-only. Stacks on
`misc/odds-and-ends-for-1p0p0`.

## Why

The library already type-checked cleanly, but the test tree did not, so the type-check target
excluded it. That left the tests free to drift and meant refactors could not lean on the checkers
for the code that exercises the library hardest.

## Changes

- Added type annotations throughout `tests/`, resolving the errors both checkers raised once the
  test tree was included.
- Made small annotation corrections in the library modules the tests pin: `_rng.py`,
  `graph/graph.py`, `partition/partition.py`, `tree/bipartition_tree.py`, `tree/spanning_tree.py`,
  `updaters/election.py`, and `updaters/tally.py`.
- Updated `Makefile` and `pyproject.toml` so the type-check target covers the tests.

## Testing

- [x] Tests pass locally
- [x] Type checks pass locally

Verified with `make type-check` (ty then pyright, over `gerrychain` and `tests`) and `make test`.
Both checkers report no errors over the combined tree.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ PYTHON_VERSION = 3.11
 VENV_DIR ?= .venv
 PKG ?= gerrychain
 TEST_PATHS ?= tests
-TYPECHECK_PATHS ?= $(PKG) tests/typing_assertions.py
+TYPECHECK_PATHS ?= $(PKG) $(TEST_PATHS)
 DOCS_PY_PATHS ?= docs/conf.py docs/_clear_notebook_outputs.py \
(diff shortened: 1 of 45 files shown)


chief-dweeb:
Not reviewed (as per your comment above)

PR #462, opened 2026-07-19
Misc/odds and ends pt2

peterrrock2:
## Summary

Don't worry about the docs failing for now.

Correctness and readability work on the balanced-edge-cut finders in `bipartition_tree.py`: renames
`one_sided_cut` to `single_district_cut`, fixes a comparison that disagreed between the two finders,
refactors the memoized finder, and finishes the `FindBalancedEdgeCutsFn` typing and docs. Stacks on
`feat/multi-member-recom`.

## Why

The two cut finders had drifted apart in a way that was easy to miss because the balanced-population
test was spelled out separately in each. The `one_sided_cut` parameter also described the cut rather
than what the flag actually controls, and the memoized finder was hard to follow (a mid-function
return and three near-identical `_Cut` constructions).

## Changes

- Renamed `one_sided_cut` to `single_district_cut` everywhere. The flag controls whether only the
  returned side must be balanced (peeling off one district) or both sides must be (bisecting); the
  new name matches what the docstrings already said.
- Fixed `_PopulatedGraph.has_ideal_population`, whose single-district branch used a strict `<` while
  its two-sided branch used `<=`, so an exactly-on-target cut at `epsilon=0` was accepted when
  bisecting but rejected when peeling. The single test is now defined once as
  `_PopulatedGraph.is_within_epsilon` and used by both finders.
- Refactored `find_balanced_edge_cuts_memoization` into a single `if`/`else` with one return, one
  `_Cut` helper, and named `below(node)`/`above(node)` sides, replacing the mid-function return and
  the duplicated construction.
- Applied `FindBalancedEdgeCutsFn` to the public `find_balanced_edge_cuts_fn` parameters, added a
  private `_BoundFindBalancedEdgeCutsFn` for the post-`partial` shape, and documented the
  degree-1 root requirement (needed by the contraction finder, not the memoized one).
- Dropped the memoized finder's root-degree filter, since any root yields the same candidate edges;
  this also lets it handle two-node trees the filter used to reject (very much an edge case).

## Testing

- [x] Tests pass locally
- [x] Added/updated tests
- [x] Type checks pass locally

Verified with `make test` (including the `--runslow` Pennsylvania freeze), `make lint`, and
`make type-check`. The memoization refactor was checked by diffing cut output (edges, subsets, and
RNG-derived weights) against the pre-refactor code across many seeds and graph shapes.

## Reviewer Notes

- **Breaking:** callers passing `one_sided_cut`, and custom cut finders accepting it, must use
  `single_district_cut`. Custom finders must now accept it unconditionally; the `inspect.signature`
  probe that made it optional has been removed. Will put in release notes.

### DIFF
--- a/gerrychain/tree/bipartition_tree.py
+++ b/gerrychain/tree/bipartition_tree.py
@@ -169,22 +169,23 @@ class _PopulatedGraph:
 
     # frm: only ever used inside this file
     #       But maybe this is intended to be used externally...
-    def has_ideal_population(self, node: Hashable, one_sided_cut: bool = False) -> bool:
+    def has_ideal_population(self, node: Hashable, single_district_cut: bool = False) -> bool:
         """Checks if a merged node is within epsilon of the ideal population.
 
         Args:
             node (Hashable): The node to check.
-            one_sided_cut (bool, optional): Whether or not we are cutting off a single district.
-                When set to False, we check if the node we are cutting and the remaining graph are
-                both within epsilon of the ideal population. When set to True, we only check if the
-                node we are cutting is within epsilon of the ideal population. Defaults to False.
+            single_district_cut (bool, optional): Whether or not we are cutting off a single
+                district. When set to False, we check if the node we are cutting and the remaining
+                graph are both within epsilon of the ideal population. When set to True, we only
+                check if the node we are cutting is within epsilon of the ideal population.
+                Defaults to False.
 
         Returns:
             bool: True if the node has an ideal population within the graph up to epsilon.
         """
 
-        if one_sided_cut:
-            return abs(self.population[node] - self.ideal_pop) < self.epsilon * self.ideal_pop
+        if single_district_cut:
+            return abs(self.population[node] - self.ideal_pop) <= self.epsilon * self.ideal_pop
 
         return (
             abs(self.population[node] - self.ideal_pop) <= self.epsilon * self.ideal_pop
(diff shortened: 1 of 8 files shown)


gerrychain/tree/bipartition_tree.py
```diff
@@ -167,29 +167,43 @@ def _contract_node(self, node: Hashable, parent: Hashable) -> None:
         self.subsets[parent] |= self.subsets[node]
         self._degrees[parent] -= 1
 
+    def is_within_epsilon(self, population: int | float) -> bool:
```
chief-dweeb, line 170: Is this an externally visible function?  If not, should we add an initial underscore?

gerrychain/tree/bipartition_tree.py
```diff
@@ -621,60 +654,49 @@ def find_balanced_edge_cuts_memoization(
...
+        )
+
+    def below(node: Hashable) -> AbstractSet[Hashable]:
+        """The subtree hanging below ``node``, including ``node`` itself."""
+        return _nodes_in_subtree(node, succ)
+
+    def above(node: Hashable) -> AbstractSet[Hashable]:
+        """Everything except the subtree below ``node``."""
+        return set(h.graph.node_indices) - _nodes_in_subtree(node, succ)
+
```
chief-dweeb, line 681: Should these have an initial underscore to indicate that they are internal functions?

gerrychain/tree/bipartition_tree.py
```diff
@@ -621,60 +654,49 @@ def find_balanced_edge_cuts_memoization(
...
-                        ),
-                        subset=frozenset(set(h.graph.node_indices) - _nodes_in_subtree(node, succ)),
-                    )
-                )
-
-        return cuts
+            if h.is_within_epsilon(tree_pop):
+                cuts.append(cut_at(node, below(node)))
+            elif h.is_within_epsilon(total_pop - tree_pop):
+                cuts.append(cut_at(node, above(node)))
```
chief-dweeb, line 692: So much clearer - thank you

chief-dweeb:
Looks just fine.

chief-dweeb (approved):
Looks fine to me...

chief-dweeb (commented):
looks good

peterrrock2:
Great! I'll work on merging this in.

The `is_within_epsilon`, `above`, and `below` functions are all used within  `find_balanced_edge_cuts_memoization`, so I counted them as part of the interface of the class. The class itself is not intended to be public, though.

PR #464, opened 2026-07-27
Documentation update to User Guide files.

chief-dweeb:
## Summary

Documentation updates to .md and .ipynb files

## Why

Purpose was to provide an alternate POV - especially to support an audience that was not expert in Python

## Changes

I intend to add comments to this PR once it is open to highlight areas where additional input is requested.
-
-
-

## Testing

- [ ] Tests pass locally
- [ ] Added/updated tests
- [ ] Manually tested relevant behavior
- [ ] Not applicable

## Reviewer Notes

-


docs/user/quickstart.ipynb
```diff
@@ -166,37 +166,41 @@
...
-    "table to the `graph` object as node attributes.\n",
+    "* The `Graph.from_json()` classmethod creates a GerryChain `Graph` of the precincts. \n",
+    "\n",
+    "    Note that this JSON file was previously created by creating a GerryChain `Graph` from a \n",
+    "    shapefile (if you don't know what a shapefile is, don't worry), and then writing that\n",
+    "    graph out as a JSON file.  We used `Graph.from_file()` on the shapefile obtained from\n",
+    "    [mggg-states/PA-shapefiles](https://github.com/mggg-states/PA-shapefiles#metadata) .\n",
+    "    By default, the `from_file()` method copies all of the data columns from the shapefile's attribute \n",
+    "    table to the `Graph` object as node attributes.  We then used `Graph.to_json()` to create\n",
+    "    the JSON file.  \n",
```
chief-dweeb, line 182: Peter - please confirm that what I have written is accurate...

docs/user/quickstart.ipynb
```diff
@@ -228,17 +232,18 @@
...
+    "And that is it! From here, you can move on to configuring and running the chain."
    ]
   },
   {
    "cell_type": "markdown",
    "id": "bcacd7fe",
    "metadata": {},
    "source": [
-    "## Running the chain\n",
+    "## Running the Chain\n",
```
chief-dweeb, line 243: The capitalization of sections and sub-sections in the User Guide is not consistent.

I was going to do a pass to fix this, but ran out of time.  Not a big deal, but also easy to fix.

docs/user/recom.ipynb
```diff
@@ -7,6 +7,16 @@
...
+    "In the previous section, we used a proposal function, `propose_random_flip`, which\n",
+    "\"flipped\" a single node from one district to another to create the next district\n",
+    "plan in the Markov chain.  This approach evolves slowly from its initial state.\n",
+    "\n",
+    "An approach which evolves much faster is to combine two adjacent districts and\n",
+    "to then randomly split the combined set of nodes into two new districts.\n",
+    "\n",
+    "The GerryChain `ReCom` class implements proposal functions that do precisely that.\n",
+    "You can see how it works in animation below:\n",
+    "\n",
```
chief-dweeb, line 19: Peter - please confirm that you are OK with this description.  Or change it if not...

docs/user/recom.ipynb
```diff
@@ -430,12 +451,24 @@
...
+    "is that `ReCom` randomizes the creation of each successive district plan by assigning\n",
+    "a random weight (between 0.0 and 1.0) to edges, and that adding a surcharge weight to an edge\n",
+    "in addition to the random weight \n",
+    "will make it *more* likely that the edge will be selected as a border edge between districts.\n",
+    "That is, *adding* weight to edges that span regions will make it more likely \n",
+    "that those edges will be chosen as border edges.  The larger the surcharge weight\n",
+    "the more likely it is that edges that span regions will be chosen as border edges.\n",
+    "\n",
+    "Let's create an ensemble of district plans using a large `region_surcharge` to keep\n",
+    "municipalities intact."
```
chief-dweeb, line 471: Peter - are you OK with this description?

docs/user/recom.ipynb
```diff
@@ -478,34 +512,104 @@
...
+   "outputs": [],
+   "source": [
+    "# Rerun the region-aware code but with a reduced region_surcharge of 0.1\n",
+    "\n",
+    "recom_chain.proposal_fn = ReCom.district_pairs_mst(\n",
+    "    pop_col=\"TOTPOP\",\n",
+    "    pop_target=ideal_population,\n",
+    "    epsilon=0.01,\n",
+    "    region_surcharge={\"muni\": 0.3},\n",
+    ")\n",
```
chief-dweeb, line 544: I added this additional example to help the user understand that different values of a region surcharge will have different effects.  

It was interesting to observe that values of 0.1 and 0.2 did not actually do a "better" job of keeping municipalities together than no region surcharge.  I think that is a result of the toy nature of our example (and the fact that the no region surcharge example happens to start by combining the first two and the last two regions which preserve municipalities).  So I went with 0.3 which nicely makes the point.

docs/user/recom.ipynb
chief-dweeb, line 642: The graphics that show region-aware plots are actually hard to follow - the reader forgets where the boundaries for the municipalities or the water districts are.

A quick fix that I am hoping that you (Peter) can make would be to include the labels for whatever is being surcharged (municipality or water district) in the middle of each block so that the reader can see when a district spans vs contains the region.

Hopefully it is easy to do, but it is certainly not a show stopper...

chief-dweeb, line 414: Not sure what was happening, but inside Jupyter Lab, this code produced a graphic with lots of controls: go to the end, go one step in either direction, play all steps, etc. while in the rendered HTML it just had a forward and backward button.

Not sure if this is a big deal or not, but certainly a bit odd...

docs/user/recom.ipynb
```diff
@@ -583,18 +688,44 @@
...
+    "a spanning tree by just cutting one edge.  All of the nodes on one side of the cut \n",
+    "edge will be connected and all of the nodes on the other side of the cut edge will \n",
+    "be connected.  This is exactly what we want ReCom to do - split a collection of nodes\n",
+    "into two connected subsets.\n",
+    "\n",
+    "ReCom works by randomly assigning weights from the interval $[0,1]$ to the \n",
+    "edges in the combined set of nodes from the two adjacent districts.  It then builds a \n",
+    "spanning tree by greedily selecting the lowest weighted edge next.  This has \n",
+    "the effect of randomizing which spanning tree is created from the \n",
+    "universe of all possible spanning trees.\n",
```
chief-dweeb, line 693: Peter - are you OK with this description?

chief-dweeb:
I only got as far as "Running a chain with Recom".

I will continue to go through the following sections of the User Guide, but I wanted to get what I have done to you as I think the sections up through Recom are the most important.

chief-dweeb (commented):
Tag you're it...

docs/user/quickstart.ipynb
peterrrock2: The workflow is correct here. There was some additional post-processing applied to the dual graph that is used in this tutorial (provenance lost to time), so users will not be able to perfectly recreate the dual graph, but I do not think that matters for this tutorial

docs/user/quickstart.ipynb
peterrrock2: Noted. I'll add this to the final polish pass that I plan to do

docs/user/recom.ipynb
peterrrock2: This works for me!

docs/user/recom.ipynb
peterrrock2: Looks good!

docs/user/recom.ipynb
peterrrock2: This is an intentional change on my part. To keep the code simple, I used the `ArtistAnimation` class that comes from matplotlib which comes with all of the extra controls. I think that a user following along with the guide should be able to figure out  what the controls do with a little bit of clicking around. It was a compromise from the previous version wherein there was a great deal of setup code that cluttered the tutorial. The additional setup did produce a nice interface that closely mirrored the rendered website documentation, but I think that it ran the risk of being overwhelming.

In the rendered documentation, I have some custom JavaScript + CSS code that simplifies the interface to the previous version so that people just reading through the documentation on the website have something more minimal to work with.

docs/user/recom.ipynb
```diff
@@ -7,6 +7,16 @@
...
+    "plan in the Markov chain.  This approach evolves slowly from its initial state.\n",
+    "\n",
+    "An approach which evolves much faster is to combine two adjacent districts and\n",
+    "to then randomly split the combined set of nodes into two new districts.\n",
+    "\n",
+    "The GerryChain `ReCom` class implements proposal functions that do precisely that.\n",
+    "You can see how it works in animation below:\n",
+    "\n",
     "<style>\n",
     "img[src$=\"gerrychain_demo.gif\"],\n",
```
peterrrock2, line 21: Note to self: put the correct gerrychain_demo gif back in here. I think I messed up a merge somewhere and put an old version in that doesn't show the edge cuts being created.

docs/user/recom.ipynb
peterrrock2: Great! I like the addition here! And 0.3 makes the point well; you are probably correct that we are not really seeing much of an effect from 0.1 or 0.2 because of how contrived our example is.

docs/user/recom.ipynb
```diff
@@ -478,34 +512,104 @@
...
+    "    epsilon=0.01,\n",
+    "    region_surcharge={\"muni\": 0.5},\n",
+    ")\n",
+    "\n",
+    "assignment_list = []\n",
+    "\n",
+    "for i, item in enumerate(recom_chain):\n",
+    "    assignment_list.append(item.assignment)"
+   ]
+  },
```
peterrrock2: I am not sure what this is adding to the tutorial. We have a code cell that is immediately overwritten by the next one. We should either animate this one as well and give a little more description, or add more comments to the next code cell to clarify what is going on.

docs/user/recom.ipynb
peterrrock2: I can probably do something about that. I have to think about what the best way to communicate that is for multiple layers. Probably best to stick to districts + one level of labels and then add a way to toggle between "water_dist" and "muni" rather than trying to cram everything in at once.

Alternatively, we can outline water dists and label munis, but I think that is the absolute limit of information density for this sort of thing.

docs/user/recom.ipynb
peterrrock2: For the most part, yes. The last paragraph is slightly inaccurate: ReCom is a class of proposal functions not all of which use Kruskal's (though that is our preferred method since it is more controllable). I would modify the first sentence of the final paragraph above to "The ReCom variant we have been using in this guide, `district_pairs_mst`, works by..."

peterrrock2 (changes requested):
Thank you for your work on this! If you want to add those quick edits in, I can start working on polishing some of the things that you have left notes on and adding the better viz for the multi-region surcharge.

docs/user/recom.ipynb
chief-dweeb: The current guide shows three graphs - one with a plan GerryChain came up with, one with the municipalities and one with the water districts.  If you changed the labels for the water districts to be letters instead of numbers (to make it easy for the reader to NOT get confused), then you could replace the three graphs with two graphs - one that showed municipality numbers and another that showed water district letters (the colors for the blocks in each graph would be the same).

I am sure you could come up with better presentations (I thought of a couple), but I would argue that less is more at this stage...  Better is the enemy of Good Enough, etc...

chief-dweeb: Cool.

I did notice that the plus sign and the minus sign did not appear to do anything, but that is definitely NOT worth spending any time on...

docs/user/recom.ipynb
chief-dweeb: Change has been made.

docs/user/recom.ipynb
chief-dweeb: Good catch - I deleted the code block and added the following text above the remaining code block that introduces a surcharge for water districts:

We reduce the surcharge for municipalities, "muni", from 0.5 to 0.2, and add
a surcharge for water districts, "water_dist", of 0.8, making preserving water districts
more important than preserving municipalities.

chief-dweeb:
Edits have been made, committed, and pushed, so go ahead and do the merge!

chief-dweeb:
Done - changes made, committed, and pushed.

Merge away...

peterrrock2:
Great! Are you still planning on dropping the migration guide stuff into this PR?

PR #468, opened 2026-07-30
Misc/add deprecation shims

peterrrock2:
## Summary

This PR restores compatibility for many GerryChain 0.3.2 call patterns whose names or callback signatures changed during the 1.0 work. Legacy calls continue to run with actionable `DeprecationWarning` messages, while the canonical 1.0 signatures remain unchanged.

The warnings identify the replacement API and state that the compatibility path will be removed in GerryChain 2.0.

Not all pre-1.0 behavior is restored, but I made sure that workflows documented in readthedocs were covered.

## What changed

- Added shared compatibility helpers in `gerrychain._deprecated` for the following:
  - renamed and ignored parameters;
  - legacy defaults;
  - renamed public functions and properties;
  - moved functions that must be resolved lazily;
  - callbacks that do not yet accept the keyword-only `rng` argument; and
  - legacy tree callback parameter names and argument ordering.
- Restored renamed parameters across partitions, updaters, constraints, optimizers, proposals, tree algorithms, and assignment helpers.
- Kept removed parameters out of the canonical signatures. Calls using those parameters are handled by wrappers and receive a warning.
- Adapted pre-1.0 proposal, acceptance, bipartition, balanced-cut, spanning-tree, and cut-choice callbacks at their shared call sites.
- Restored moved or renamed public entry points in `gerrychain.tree`, `gerrychain.grid`, `gerrychain.updaters`, and `gerrychain.constraints`.
- Reverted to `L1` as the canonical spelling for the compactness helpers (in hindsight `L_1` was confusing when the proper symbol is $L^1$ and not $L_1$)
- Restored deprecated attribute access for:
  - `MarkovChain.proposal`, `.accept`, `.initial_state`, and `.is_valid`; and
  - `.func` on all Bounds classes.

## Compatibility behavior

- Supplying a legacy and canonical parameter name together raises a clear `TypeError`.
- Legacy callbacks are inspected and adapted once before use. Exceptions raised inside a callback are not caught and retried.
- Canonical 1.0 calls do not emit deprecation warnings.
- Deprecated calls identify the old name, its replacement, and the planned 2.0 removal.

## Tests

- Added focused tests for renamed parameters, aliases, ignored arguments, callback adaptation, warning messages, canonical signatures, and legacy read/write attributes.
- Full test suite: `533 passed, 9 skipped, 2 xfailed`.

## Exclusions

This PR did not restore:

- the pre-1.0 `Graph` API or other behavior made obsolete by the RustworkX migration;
- the old `ReCom` interface;
- old positional layouts where parameters were reordered or inserted; or
- global `random.seed(...)` control over the new explicit RNG streams;

## Reviewer Notes

- Thorough review not required. A quick sanity check to make sure I haven't missed anything substantial is appreciated.

### DIFF
--- a/gerrychain/_deprecated.py
+++ b/gerrychain/_deprecated.py
@@ -205,17 +205,28 @@ def allow_legacy_missing_rng(fn: Callable[P, R]) -> Callable[P, R]:
     return wrapped
 
 
-def deprecated_property(old_name: str, new_name: str) -> property:
+def deprecated_property(old_name: str, new_name: str, *, writable: bool = False) -> property:
     """Return a property forwarding a deprecated attribute name to its replacement."""
 
-    def get(instance: object) -> Any:
+    def warn() -> None:
         _warn(
             f"{old_name} is deprecated; use {new_name} instead. The legacy name will be removed "
-            "in GerryChain 2.0."
+            "in GerryChain 2.0.",
+            stacklevel=4,
         )
+
+    def get(instance: object) -> Any:
+        warn()
         return getattr(instance, new_name)
 
-    return property(get)
+    if not writable:
+        return property(get)
+
+    def set_(instance: object, value: Any) -> None:
+        warn()
+        setattr(instance, new_name, value)
+
+    return property(get, set_)
 
 
 def deprecated_lazy_alias(
--- a/gerrychain/_deprecated.py
+++ b/gerrychain/_deprecated.py
@@ -1,0 +1,301 @@
+"""Compatibility helpers for APIs deprecated after GerryChain 0.3.2."""
+
+from __future__ import annotations
+
+import functools
+import importlib
+import inspect
+import random
+import warnings
+from collections.abc import Callable, Collection, Mapping
+from typing import Any, ParamSpec, TypeVar, cast
+
+P = ParamSpec("P")
+R = TypeVar("R")
+
+
+def _warn(message: str, stacklevel: int = 3) -> None:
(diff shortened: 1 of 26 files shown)


gerrychain/_deprecated.py
```diff
@@ -0,0 +1,481 @@
...
+                        f"{fn_name}() omitted {name!r}; using the legacy default {value!r}. "
+                        f"Pass {name}=... explicitly. The implicit default will be removed in "
+                        "GerryChain 2.0."
+                    )
+
+            return fn(*args, **mutable_kwargs)
+
+        return wrapped
+
+    return decorate
```
chief-dweeb, line 111: This is magic.

Kudos!

chief-dweeb (approved):
This is great - especially for a major release.

It also makes the change-log less important, as users will discover what has changed in a very natural and convenient way.

I do wonder, however, how you identified all of the changes that would need to be dealt with.  For my change-log, I used the ast module to go over both codebases and then compare the two in code.  What did you do?

peterrrock2:
> This is great - especially for a major release.
> 
> It also makes the change-log less important, as users will discover what has changed in a very natural and convenient way.
> 
> I do wonder, however, how you identified all of the changes that would need to be dealt with. For my change-log, I used the ast module to go over both codebases and then compare the two in code. What did you do?

Thank you! I used the super scientific "run the old tutorial guide and some old scripts and put a shim around anything that breaks" method to find all of these. I figure that I have used more features of GerryChain than most, so anything I bumped into was worth shimming.

